PR #10, opened 2020-05-07
Added support for viaplay

MrMamen:
Currently works only with viaplay.no, It is available with presumably same API in other countries but with different top-level domain.

### DIFF
--- a/src/modules/history/streaming-services/viaplay/ViaplayApi.ts
+++ b/src/modules/history/streaming-services/viaplay/ViaplayApi.ts
@@ -1,0 +1,149 @@
+import * as moment from 'moment';
+import { TraktSearch } from '../../../../api/TraktSearch';
+import { TraktSync } from '../../../../api/TraktSync';
+import { Item } from '../../../../models/Item';
+import { Errors } from '../../../../services/Errors';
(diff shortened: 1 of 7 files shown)


src/modules/history/streaming-services/viaplay/ViaplayApi.ts
```diff
@@ -0,0 +1,149 @@
...
+      Errors.error('Failed to load Viaplay history.', err);
+      await EventDispatcher.dispatch(Events.STREAMING_SERVICE_HISTORY_LOAD_ERROR, { error: err });
+    }
+  }
+
+  parseHistoryItem(historyItem: ViaplayProduct): Item {
+    let item: Item = null;
+    const year = historyItem.content.production.year;
+    const percentageWatched = historyItem.user.progress?.elapsedPercent || 0;
+    const watchedAt = moment(historyItem.user.progress?.updated);
```
MrMamen, line 93: I found one item that had no progress-field in my own history. Not quite sure what do with such items. Right now this will transform into today's date. Another approach is to remove the item entirely.

rafaelgomesxyz:
Oh, I forgot about this, my bad. Things have been a bit busy recently, but my semester ends in a couple weeks, so I'll have more time to focus on the extension.

PR #12, opened 2020-06-21
Setup tools for consistent style, linting and type checking

### LINKED ISSUE #5: Define style guide
Setup prettier and/or lint --fix on commit for consistent code styles.

https://eslint.org/
https://prettier.io/
https://github.com/typicode/husky

### LINKED ISSUE #15: Typescript types be defined where they belong
typedefs.d.ts currently holds all types. This should be avoided in the future as it grows longer, can cause merge conflicts, and most of them clearly belong to an specific streaming service. They should instead be defined inline (they can still be exported) or in the same directory as the files accessing them. They should not be global if possible, but rather be imported as needed.

rafaelgomesxyz:
Fixes #5 

Don't merge yet, I'm still working on some stuff. I just wanted your feedback so far, to see if you agree with the current rules (it's basically just using the recommended config from all the plugins for now). I considered the Airbnb style, but I think it's too strict.

The difference between the `lint*` and the `lint*typed` scripts is that the `lint*` scripts don't compile the project with TypeScript (which is faster, but with no type-checking). This is recommended by https://github.com/typescript-eslint/typescript-eslint/tree/master/packages/eslint-plugin to separate the flow and still take advantage of type-checking. So basically we should lint with no type-checking while working on the code, because it's faster, and when we're ready to commit we lint with type-checking. This is why the pre-commit hook is configured to lint with type-checking.

I'm going over some of the non-recommended rules from the plugins to see if there are any other rules that could be relevant for us. If you have any suggestions, let me know. Something that I want to enforce is using this:

```
classMethod = () => {
};
```

Over this:

```
constructor() {
  this.classMethod = this.classMethod.bind(this);
}

classMethod() {
}
```

As the former method is much simpler and doesn't require binding because of the arrow function. I'm not sure if there's a rule for this, though.


rafaelgomesxyz:
Oh, didn't know about drafts. Leaving it like this for now.

package.json
```diff
@@ -1,68 +1,106 @@
...
+		"webpack-cli": "^3.3.10"
+	},
+	"husky": {
+		"hooks": {
+			"pre-commit": "lint-staged"
+		}
+	},
+	"lint-staged": {
+		"*.json": "prettier --write",
+		"*.{js,jsx,ts,tsx}": [
```
MrMamen, line 104: prettier should run on these files as well?

MrMamen (commented):
I think "Lint staged" should go through .md, .html, .json files as well. Seems to be that they have been handles in the PR, but not in the scripts?

MrMamen:
I don't really have any preferences on the actual rules. I have a habbit of using double quotes, but since the linter makes everything the same way it donesn't really matter to me.

.eslintrc.typed.js
```diff
@@ -0,0 +1,45 @@
+module.exports = {
```
MrMamen, line 1: What's the point of having two lint-config-files? Seems to me that there are typescript rules in both?

package.json
rafaelgomesxyz: It does. `eslint-plugin-prettier` shows Prettier errors as ESLint errors and automatically fixes them when run with `--fix`. So when you run ESLint, it's automatically running Prettier as well. That's why I separated .json above, because ESLint can't run on .json. Here's how we configure the plugin: https://github.com/trakt-tools/universal-trakt-sync/blob/prettier-eslint/.eslintrc.js#L13

rafaelgomesxyz:
> I think "Lint staged" should go through .md, .html, .json files as well. Seems to be that they have been handles in the PR, but not in the scripts?

Right, I forgot to include them in the pre-commit hook. I did apply Prettier to the entire project, which is why everything is formatted.

> I don't really have any preferences on the actual rules. I have a habbit of using double quotes, but since the linter makes everything the same way it donesn't really matter to me.

I prefer single quotes because it's just one key stroke, while double quotes require pressing Shift. But `lint-staged` leaves your local files alone, so you can always use double quotes locally.

The only thing that directly affects local code is EditorConfig, but only if your editor supports it (on VS Code it only works if you install the plugin). And the only really controversial rule in it is the enforcing of tabs instead of spaces (I prefer tabs because each developer can use their preferred indentation width). So if you prefer spaces, you can just disable it.

Edit: Actually, looks like `lint-staged` doesn't leave local files alone. I was under the impression it did...

.eslintrc.typed.js
rafaelgomesxyz: It's what I mentioned in the OP about linting with type-checking. `.eslintrc.js` is the normal linting, which is faster. `.eslintrc.typed.js` is the linting with type-checking. It compiles the entire project with TypeScript and uses the type information to apply the rules. The difference is pretty much here: https://github.com/trakt-tools/universal-trakt-sync/blob/prettier-eslint/.eslintrc.typed.js#L33 This allows for a more robust linting, but it's slower, hence why it's only applied in the pre-commit hook.

rafaelgomesxyz:
Ok, I think I'm done here. I found a plugin that does what I wanted (enforce arrow functions, so we don't have to bind stuff): https://github.com/TristonJ/eslint-plugin-prefer-arrow But it's not quite working for class methods, so I made a PR over there. Until the PR is merged, I'm including the fix here in the `eslint-plugin-prefer-arrow` folder.

rafaelgomesxyz:
Also, I think we've been merging PRs. Wouldn't it be best to squash and merge to keep a clean history?

MrMamen:
> Also, I think we've been merging PRs. Wouldn't it be best to squash and merge to keep a clean history?

For this PR where all files are touched a squash would be preferred by me. In other branches it depends. At least if several developers has worked on a branch I would prefer to keep commits intact.

MrMamen (approved):

rafaelgomesxyz:
> At least if several developers has worked on a branch I would prefer to keep commits intact.

If what you're worried about is developers not receiving contributor credit, I'm pretty sure GitHub has fixed that recently, so that all commit authors receive credit. See https://github.blog/changelog/2019-12-19-improved-attribution-when-squashing-commits/ and https://github.com/isaacs/github/issues/1303#issuecomment-595595284.

MrMamen:
> > At least if several developers has worked on a branch I would prefer to keep commits intact.
> 
> If what you're worried about is developers not receiving contributor credit, I'm pretty sure GitHub has fixed that recently, so that all commit authors receive credit. See https://github.blog/changelog/2019-12-19-improved-attribution-when-squashing-commits/ and [isaacs/github#1303 (comment)](https://github.com/isaacs/github/issues/1303#issuecomment-595595284).

That's great! Might still be a good idea to consider keeping individual commits due to see blame history, but not really important.

PR #17, opened 2020-07-14
Finish converting files to TypeScript

rafaelgomesxyz:
Will be rebased after #12 is merged.

PR #18, opened 2020-07-14
Make streaming services optional

### LINKED ISSUE #9: Make streaming services optional
Remove mandatory permissions for each service and make them optional. Each permission is then required if the user enables sync for the corresponding service.

rafaelgomesxyz:
Fixes #9 

Will be rebased after #12, #17 and #20 are merged.

This also makes it much easier to add new streaming services, since it removes the need to create the `*Store` and `*Page` files. All you have to do now is create the `*Api` file, import this file in `pages.ts`, and add the service to `streaming-services.ts`.


MrMamen:
I'll have another look at this when it's rebased.

rafaelgomesxyz:
Rebased. I probably should have made smaller commits to make it easier to review. I'll do that from now on.

rafaelgomesxyz:
Can you translate the Norwegian file for the new messages?

src/streaming-services.ts
```diff
@@ -0,0 +1,29 @@
...
+	nrk: {
+		id: 'nrk',
+		name: 'NRK',
+		homePage: 'https://tv.nrk.no/',
+		hostPattern: '*://*.tv.nrk.no/*',
+	},
+	viaplay: {
+		id: 'viaplay',
+		name: 'Viaplay',
+		homePage: 'https://viaplay.no/',
```
MrMamen, line 26: Not relevant for this branch, but have to figure out a way to enable viaplay from other domains, such as .se, .fi, .dk, etc.

rafaelgomesxyz: Since `hostPattern` is responsible for defining the domain, we could just change it to an array of domains instead.

MrMamen (approved):
Seems good. As a suggestion I would consider an "add all" switch.

rafaelgomesxyz:
Added the buttons and also the option to use multiple domains.

MrMamen:
I think you jinxed it in one of the latest commits:

`TypeError: Object.values(...).map(...).flat(...) is not a function or its return value is not iterable
    at getManifest (\universal-trakt-sync\webpack.config.ts:189:6)
    at runFinalSteps (\universal-trakt-sync\webpack.config.ts:262:10)
    at plugins.env.test.plugins.runAfterBuild (\universal-trakt-sync\webpack.config.ts:144:57)
`

rafaelgomesxyz:
Weird, here it's working fine. That's during build, right? I built both in dev and production mode and got no errors.

rafaelgomesxyz:
I think I got it, `flat()` is only available since Node 12 (https://node.green/#ES2019-features-Array-prototype--flat--flatMap-). I assume you're on an older version?

We should probably change `.nvmrc` to 12.5.0, which is the oldest version compatible with the newest JavaScript features.

MrMamen:
Yup quite right. Got it working with node 12. nvmrc should be upgraded yes. LTS version is always a safe choice I think.

PR #23, opened 2020-07-17
Merge with scrobbler

### LINKED ISSUE #19: Merge with scrobbler or third repo with common code?
There's a lot of code that will be duplicated between the scrobbler and the sync extensions. In order to reduce that, I can think of two solutions:

* Merge both extensions (it's not like the size is that much bigger, and users might prefer to have a single extension installed rather than multiple ones)
* Create a third repo to host common code (which we then publish as a package and use as a dependency for each extension)

What do you think?

rafaelgomesxyz:
Fixes #19 

Still need to merge the actual scrobblers, but most of it is done.


src/_locales/en/messages.json
```diff
@@ -15,7 +15,7 @@
 		"message": "Add with release date."
 	},
 	"aboutMessage": {
-		"message": "Sync the history from your favorite streaming services to Trakt.tv!"
+		"message": "Scrobble / sync the history from your favorite streaming services to Trakt.tv!"
```
MrMamen: Could we simplify and just say: "Send your activity from your favorite streaming services to Trakt.tv"?

src/components/UtsLeftRight.tsx
```diff
@@ -5,7 +5,7 @@ interface UtsLeftRightProps {
 	centerVertically: boolean;
 	className?: string;
 	left: React.ReactElement;
-	right?: React.ReactElement;
+	right?: React.ReactElement | null;
```
MrMamen: Does it have to be optional if it can be null?

src/modules/background/background.ts
```diff
@@ -1,33 +1,75 @@
...
-import { Tabs } from '../../services/Tabs';
-import { streamingServices } from '../../streaming-services';
+import { streamingServices } from '../../streaming-services/streaming-services';
 
-interface MessageRequest {
-	action: 'check-login' | 'finish-login' | 'login' | 'logout' | 'send-request';
-	url: string;
+export type MessageRequest =
+	| {
+			action:
```
MrMamen: Probably why a lot of people don't like TypeScript, but I prefer defining every single one of these as Interfaces. It will be a bit more overhead, but I think think it looks cleaner.

src/_locales/en/messages.json
rafaelgomesxyz: Yup

src/components/UtsLeftRight.tsx
rafaelgomesxyz: You're right, it's better to leave it just as optional, since this allows us to omit the property entirely if we don't need it.

src/modules/background/background.ts
rafaelgomesxyz: I agree. It also makes it easier in the future if we need to add a property to an action.

src/services/Events.ts
```diff
@@ -5,70 +5,121 @@ import { StreamingServiceId } from '../streaming-services/streaming-services';
 import { StorageValuesOptions, StorageValuesSyncOptions } from './BrowserStorage';
 import { Errors } from './Errors';
 import { RequestException } from './Requests';
-
-export enum Events {
```
rafaelgomesxyz, line 7: I don't see the benefit of using an enum here, is there one? Using a conditional type makes it much easier to do generic typings. If you think an enum is best, I can revert.

rafaelgomesxyz:
Alright, think I'm done here.

rafaelgomesxyz:
The only bug I'm experiencing is that sometimes the popup doesn't load, which appears to be the same bug that affects traktflix (https://github.com/tegon/traktflix/issues/100). But after about 3 times of clicking the button, it loads. I'm trying to see what's causing this.

**Edit:** Actually, doesn't look like it's that issue, but I was sure this also happened in traktflix.

MrMamen (changes requested):
I get an error when toggling: Show browser notifications. (Failed to save option.)

MrMamen:
> I get an error when toggling: Show browser notifications. (Failed to save option.)

Same with Netflix. Not able to disable it.

rafaelgomesxyz:
> I get an error when toggling: Show browser notifications. (Failed to save option.)

For notifications, it's because I forgot to add the `notifications` permission to the manifest, but I don't know why you can't disable Netflix. I'll test with a clean profile.

MrMamen:
> > I get an error when toggling: Show browser notifications. (Failed to save option.)
> 
> For notifications, it's because I forgot to add the `notifications` permission to the manifest, but I don't know why you can't disable Netflix. I'll test with a clean profile.

Probably related but I'm not able to scrobble from Netflix either. I did try to clean all cache & data, but same problem.

src/_locales/pt_BR/messages.json
```diff
@@ -66,6 +66,15 @@
 	"correctWrongItemFailed": {
 		"message": "Ocorreu um erro ao corrigir o item errado."
 	},
+	"couldNotScrobble": {
+		"message": "Não foi possível realizar o scrobble."
+	},
+	"disableScrobblingDescription": {
+		"message": "With this option enabled, the extension will not automatically scrobble what you are watching. This is useful if you only want to use the history sync."
```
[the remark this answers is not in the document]
rafaelgomesxyz: Hehe, coincidentally I've actually added an eslint plugin for translation files to the hbo-go branch, and I modified it to check for duplicate messages (https://github.com/godaddy/eslint-plugin-i18n-json/pull/33), which can show cases like this as errors, so we never forget to translate messages again. \o/

rafaelgomesxyz:
> Probably related but I'm not able to scrobble from Netflix either. I did try to clean all cache & data, but same problem.

Ok, I can reproduce it on Chromium.

**Edit:** This is a bizarre error. It looks like the extension still thinks that Netflix is a required permission. I tried completely wiping the extension from my Chromium folder and it still thinks that.

rafaelgomesxyz:
Ok, I fixed the scrobble bug.

As for not being able to disable Netflix, I think this is because we're loading the extension as unpacked, so it considers all permissions as required. I don't think this should happen with an actual extension coming from the store.

MrMamen:
> Ok, I fixed the scrobble bug.
> 
> 
> 
> As for not being able to disable Netflix, I think this is because we're loading the extension as unpacked, so it considers all permissions as required. I don't think this should happen with an actual extension coming from the store.

But why are the others toggable?

rafaelgomesxyz:
> > Ok, I fixed the scrobble bug.
> > As for not being able to disable Netflix, I think this is because we're loading the extension as unpacked, so it considers all permissions as required. I don't think this should happen with an actual extension coming from the store.
> 
> But why are the others toggable?

They're not toggable for me. Probably because I erased the extension and added it again.

MrMamen:
> > > Ok, I fixed the scrobble bug.
> > > As for not being able to disable Netflix, I think this is because we're loading the extension as unpacked, so it considers all permissions as required. I don't think this should happen with an actual extension coming from the store.
> > 
> > 
> > But why are the others toggable?
> 
> They're not toggable for me. Probably because I erased the extension and added it again.

Just for the record. The services with scrobbling are not toggable for me, but the ones with just history sync are.

PR #24, opened 2020-07-19
Add HBO Go sync

rafaelgomesxyz:
Still need to work on some things, but if you already want to translate the missing messages (it's really easy to check for them now, you just need to run `npm run lint-translations`, since it should show errors if there are duplicate values compared to the English messages).

The local eslint plugin will only be here until my PRs are merged in the plugin repos (if that takes too long, I'll fork them and upload the packages myself).


rafaelgomesxyz:
I've separated the translations lint to make things easier. So run `npm run lint-translations` to detect missing translations.

rafaelgomesxyz:
For reference, here's how the missing watched date dialog looks:

MrMamen:
> For reference, here's how the missing watched date dialog looks:
> 

Nice! I need this feature in one of my services as well. Just one random entry that had a missing date. Very strange. I just put today's date as a workaround.

rafaelgomesxyz:
> Nice! I need this feature in one of my services as well. Just one random entry that had a missing date. Very strange. I just put today's date as a workaround.

Hm, we should probably add a third option to use current date, so the user doesn't have to select it in the date picker.

PR #26, opened 2020-08-08
Update messages.json

MrMamen:


rafaelgomesxyz:
The checks failed because of 'add'. If that's actually the translation, then we can add the key to the ignored keys in https://github.com/trakt-tools/universal-trakt-scrobbler/blob/master/.eslintrc.js#L26.

rafaelgomesxyz (approved):

PR #30, opened 2020-08-26
ItemCard is stuck in loading animation

MrMamen:


src/api/TraktSearch.ts
```diff
@@ -104,6 +104,7 @@ class _TraktSearch extends TraktApi {
 			await EventDispatcher.dispatch('SEARCH_SUCCESS', null, { searchItem });
 		} catch (err) {
 			await EventDispatcher.dispatch('SEARCH_ERROR', null, { error: err as Error });
+			throw err;
```
MrMamen, line 107: This is needed so that loadTraktItemHistory sets item.trakt = null in the catch-block in Api.ts.

rafaelgomesxyz (approved):

PR #33, opened 2020-09-06
Add option to send / receive suggestions for wrong items

### LINKED ISSUE #14: Add option to correct wrong items
- [x] Option to correct locally (#22)
- [x] Option to submit corrections to a Google sheet, so other users can receive suggestions (#33)

rafaelgomesxyz:
This finally closes #14


package.json
```diff
@@ -14,20 +14,20 @@
 		"generate-release": "node ./scripts/generateRelease.js",
 		"start": "webpack --env.development --env.watch",
 		"zip": "gulp zip",
-		"prettier": "prettier '**/*.{js,jsx,ts,tsx}'",
+		"prettier": "prettier \"**/*.{js,jsx,ts,tsx}\"",
```
MrMamen, line 17: Still not able to match anything on windows of some strange reason. (husky / lint-staged still works though)

src/api/TraktScrobble.ts
```diff
@@ -41,7 +41,7 @@ class _TraktScrobble extends TraktApi {
 	}
 
 	start = async (item: TraktItem): Promise<void> => {
-		if (!Shared.isBackgroundPage) {
+		if (Shared.pageType !== 'background') {
```
MrMamen, line 44: I like that better.

package.json
MrMamen: Sorry I was mistaken. Now it works!

MrMamen (commented):
I get this error when trying to access the options-page:
`options.js:91991 Uncaught TypeError: Cannot read property 'name' of undefined
    at StreamingServiceOption (options.js:91991)
    at renderWithHooks (options.js:72730)
    at mountIndeterminateComponent (options.js:75409)
    at beginWork (options.js:76523)
    at HTMLUnknownElement.callCallback (options.js:58115)
    at Object.invokeGuardedCallbackDev (options.js:58164)
    at invokeGuardedCallback (options.js:58219)
    at beginWork$1 (options.js:81130)
    at performUnitOfWork (options.js:80084)
    at workLoopSync (options.js:80057)`

rafaelgomesxyz:
Odd, I can open the options page just fine.

MrMamen:
Oh. This is due to the fact that I worked on another service on another branch. BrowserStorage I guess. While there is nothing wrong with your branch it does mean that we can create deadlocked situations which we have to keep in mind for future updates.

MrMamen (dismissed):
I've been able to test this now. Seems to work good, except that it seems a bit slow. I guess this might be because of google fomrs api?

rafaelgomesxyz:
Yeah, Google's API is a bit slow, since it's free. Ideally we would need our own server. I have one that I use for my other project, but I'm not sure if it would be able to handle it either, since I have the cheapest tier.

MrMamen:
Just leave it. Using a public service has some advantages. Should the URL for the form be public somehow?

rafaelgomesxyz:
Yeah, I'll put the URL in the options page like on traktflix, but I'll need to make it read-only. Do you have a Gmail so I can give you write permission on it?

MrMamen:
yeah
erikandre.mamen@gmail.com

– Erik-André Mamen

‐‐‐‐‐‐‐ Original Message ‐‐‐‐‐‐‐

src/common/BrowserStorage.ts
```diff
@@ -37,6 +38,11 @@ export type CorrectItem = {
 	url: string;
 };
 
+export type ScrobblingItem = Omit<ItemBase, ''> & {
```
MrMamen, line 42: I don't get this Omit-pattern. As nothing is omitted, isn't ScrobblingItem the same as ItemBase except for the extension at the end?

rafaelgomesxyz: If you don't do it like this, `browser.storage.local.set` doesn't like it, for some reason. I think it's because it's an interface, and the empty Omit changes it to an actual object?

rafaelgomesxyz:
Still a couple things left to do here:

- [x] Fix the design of the correction dialog in the popup (since it's too small, things overlap)
- [x] Cache suggestions (so we don't need to keep making the same requests to the Google Sheet)

rafaelgomesxyz:
Alright, I think I'm done here.

MrMamen (approved):

PR #35, opened 2020-09-08
Add option to use dark theme

rafaelgomesxyz:


MrMamen (dismissed):
Nice! You're on a roll!
If I were to give some additional tips, it would be to have a lighter theme for the pop-up-window in light mode.
Also a "follow system default" could be a nice addition.

src/components/ThemeWrapper.tsx
```diff
@@ -0,0 +1,98 @@
...
+		value: 'system',
+		palette: 'light',
+	});
+	const prefersLightMode = useMediaQuery('(prefers-color-scheme: light)', { noSsr: true });
+	const systemPalette = prefersLightMode ? 'light' : 'dark';
+
+	React.useEffect(() => {
+		const setTheme = async () => {
+			const { options } = await BrowserStorage.get('options');
+			const themeValue = options?.theme ?? 'light';
```
MrMamen: perhaps system should be default?

MrMamen (dismissed):
Visited links are slightly difficult to read on dark theme, but not a major issue.

src/components/ThemeWrapper.tsx
rafaelgomesxyz: Oops, it was supposed to be. Will fix.

rafaelgomesxyz:
> Visited links are slightly difficult to read on dark theme, but not a major issue.

Probably because I was using an anchor instead of a Link component. It should be a bit better now.

MrMamen (approved):
Well done. I think this is ready for merging now.

Issue #21, opened 2020-07-15
Add images


rafaelgomesxyz:
I'm thinking they should be added as backgrounds in the cards, with some transparency, to save space.

PR #37, opened 2020-09-16
new NRK API

### LINKED ISSUE #36: NRK history sync stopped working
This seems to be because of an API redesign by NRK.

MrMamen:


src/streaming-services/nrk/NrkApi.ts
```diff
@@ -65,23 +129,29 @@ class _NrkApi extends Api {
...
+						Authorization: 'Bearer ' + this.token,
+					},
 				});
-				const responseJson = JSON.parse(responseText) as NrkHistoryItem[];
-				if (responseJson && responseJson.length > 0) {
-					itemsToLoad -= responseJson.length;
-					historyItems.push(...responseJson);
+				const responseJson = JSON.parse(responseText) as NrkProgressResponse;
+				const { progresses } = responseJson;
+				this.HISTORY_API_URL = this.API_HOST_URL + responseJson._links.next.href;
```
MrMamen: Since this is an API with navigational features, do you think this is the correct way to achieve navigation? The actual URL is not easy to replicate as it consists of a key-param, which functions as the starting point.

MrMamen:
It doesn't seem to load images as #21 , not sure why.

rafaelgomesxyz:
Will review today, sorry for the delay.

src/streaming-services/nrk/NrkApi.ts
```diff
@@ -100,51 +170,63 @@ class _NrkApi extends Api {
...
-		const year = program.productionYear;
-		const percentageWatched = parseInt(historyItem.lastSeen.percentageWatched, 10);
-		const watchedDate = this.convertAspNetJSONDateToDateObject(historyItem.lastSeen.at);
-		const watchedAt = watchedDate ? moment(watchedDate) : undefined;
+		const id = historyItem.id;
+		const programInfo = historyItem._embedded.programs;
+		const programPage = await this.lookupNrkItem(programInfo._links.self.href);
+		const type = programPage._links.seriesPage !== undefined ? 'show' : 'movie';
+		const titleInfo = programInfo.titles;
+		const title = programPage.moreInformation.originalTitle ?? titleInfo.title;
```
MrMamen, line 183: When testing I've figured out that sometimes the title would have been found if the title from titleInfo was used instead. A bit annoying since it seems safer to go with originalTitle, since translations is often missing on trakt.
originalTitle sometimes includes seasons, and even episode title, but it's rather rare, so not sure if it's worth making special code for it.

rafaelgomesxyz: Perhaps we should use a fallback system in this situation? Try first with `originalTitle` and if no results are found try with the other one?

To make this more generic, we could add a new field to `Item` called `alternativeTitles` or something, and this would be an array that is tried, in order, if the main title didn't match.

rafaelgomesxyz: And then all we would need to do is modify `TraktSearch.find` to do a loop that only returns when an item is found or when it goes through all titles.

MrMamen: This seems like a good idea, but not really a job for this PR.

rafaelgomesxyz: Yeah, I know, I was just throwing it out there.

src/streaming-services/nrk/NrkApi.ts
rafaelgomesxyz: If the URL is constantly changing, I think this is ok. It might be preferable to rename it as `historyApiUrl` just to separate the notion of constants and variables, so we know that this value will change, but I don't think it's that important.

rafaelgomesxyz: This actually leads to an error for me. I guess because my account is new and I only have 2 items in my history, `responseJson._links.next` isn't available in the response.

src/streaming-services/nrk/NrkApi.ts
```diff
@@ -65,23 +129,29 @@ class _NrkApi extends Api {
...
+					},
 				});
-				const responseJson = JSON.parse(responseText) as NrkHistoryItem[];
-				if (responseJson && responseJson.length > 0) {
-					itemsToLoad -= responseJson.length;
-					historyItems.push(...responseJson);
+				const responseJson = JSON.parse(responseText) as NrkProgressResponse;
+				const { progresses } = responseJson;
+				this.HISTORY_API_URL = this.API_HOST_URL + responseJson._links.next.href;
+				if (progresses.length > 0) {
```
rafaelgomesxyz, line 148: I think you mentioned this before, but this causes and endless loop because this condition will always be true (unless the user has 0 items in their history), so it keeps loading the same page over and over again.

To fix this, we should probably keep a list of the progress IDs that have already been fetched, and check for new ones when loading a new page.

rafaelgomesxyz (changes requested):
Images appear to be working fine for me, but I only tested one show. Can you tell me what show doesn't show images for you? Or is it all of them?

src/streaming-services/nrk/NrkApi.ts
MrMamen: I think this is no longer the case?

MrMamen:
> Images appear to be working fine for me, but I only tested one show. Can you tell me what show doesn't show images for you? Or is it all of them?
Two examples:
https://tv.nrk.no/serie/blindpassasjer/sesong/1/episode/1
https://tv.nrk.no/serie/foerstegangstjenesten/sesong/1/episode/1

src/streaming-services/nrk/NrkApi.ts
```diff
@@ -140,7 +140,11 @@ class _NrkApi extends Api {
 				});
 				const responseJson = JSON.parse(responseText) as NrkProgressResponse;
 				const { progresses } = responseJson;
-				this.HISTORY_API_URL = this.API_HOST_URL + responseJson._links.next.href;
+				if (responseJson._links.next) {
```
rafaelgomesxyz, line 143: Right, I hadn't thought of checking for the next URL to make sure we're on the last page. So we don't need to store the progress IDs after all.

src/streaming-services/nrk/NrkApi.ts
```diff
@@ -65,23 +129,33 @@ class _NrkApi extends Api {
...
-				const responseJson = JSON.parse(responseText) as NrkHistoryItem[];
-				if (responseJson && responseJson.length > 0) {
-					itemsToLoad -= responseJson.length;
-					historyItems.push(...responseJson);
+				const responseJson = JSON.parse(responseText) as NrkProgressResponse;
+				const { progresses } = responseJson;
+				if (responseJson._links.next) {
+					this.HISTORY_API_URL = this.API_HOST_URL + responseJson._links.next.href;
+				} else {
+					isLastPage = true;
```
rafaelgssa, line 146: This is now redundant, since we're already setting the last page based on the next URL.

src/streaming-services/nrk/NrkApi.ts
```diff
@@ -65,23 +129,33 @@ class _NrkApi extends Api {
...
+				}
+				if (progresses.length > 0) {
+					itemsToLoad -= progresses.length;
+					historyItems.push(...progresses);
 				} else {
 					isLastPage = true;
```
rafaelgomesxyz: This is now redundant, since we're already setting the last page based on the next URL.

rafaelgomesxyz:
> > Images appear to be working fine for me, but I only tested one show. Can you tell me what show doesn't show images for you? Or is it all of them?
> > Two examples:
> > https://tv.nrk.no/serie/blindpassasjer/sesong/1/episode/1
> > https://tv.nrk.no/serie/foerstegangstjenesten/sesong/1/episode/1

Weird, I'm getting images for those. It might be a cache thing.

rafaelgomesxyz (dismissed):

MrMamen:
> > > Images appear to be working fine for me, but I only tested one show. Can you tell me what show doesn't show images for you? Or is it all of them?
> 
> > > Two examples:
> 
> > > https://tv.nrk.no/serie/blindpassasjer/sesong/1/episode/1
> 
> > > https://tv.nrk.no/serie/foerstegangstjenesten/sesong/1/episode/1
> 
> 
> 
> Weird, I'm getting images for those. It might be a cache thing.

Hmm. Do i need to set some token or id to be able to update the drive-sheet? These values are now in the sheet, but not when I tested.

rafaelgomesxyz:
You shouldn't need anything, no. If you get it again, can you check the requests to the sheet to see what they're sending / receiving?

rafaelgomesxyz:
Could also be that you haven't given permission to the sheet? Do you have the option to submit corrections enabled?

rafaelgomesxyz (approved):

MrMamen:
The only option I see is to allow corrections to be sent, and the correction sheet works. I seems to fetch images that already are in the sheet, but not new ones perhaps?

MrMamen:
BTW, I still don't see the results that you were able to see. This might be what you want to debug:
https://script.google.com/macros/s/AKfycbyz0AYx9-R2cKHxyyRNrMYbqUnqvJbiYxSZTFV0/exec?serviceId=nrk&ids=13ADA1790E5BB6F0B5D03C30828C38C2,D21296E8803D315C4CE7BA959C437931,9D5F0930E4F76DC1E49C0EDBE62270A4

It returns a 302, but has no data in return.

The correction script does work however.

rafaelgomesxyz:
That's the thing, that link returns the data for me. I wonder if it's a regional thing? I'll test with a VPN.

MrMamen:
No wait. This is for the corrections?

rafaelgomesxyz:
Ah, you're right, it's for corrections.

MrMamen:
Right. I get the results. But was confused with a redirect where I get the results. But I get no entry for images I think. Strange.

PR #39, opened 2020-09-24
Mismatch with displayed and sent data.

### LINKED ISSUE #38: Displayed item doesn't match what is actually sent to Trakt
Due to a bug, sometimes what is shown as a trakt match, doesn't match what is actually sent to Trakt when syncing. This is due to a merging of two different searches (show & episode) that might not match.

MrMamen:


package.json
```diff
@@ -162,7 +162,8 @@
 	"prettier": {
 		"printWidth": 100,
 		"useTabs": true,
-		"singleQuote": true
+		"singleQuote": true,
+		"endOfLine": "auto"
```
MrMamen: Because of windows-trouble

MrMamen:
The current implementation will mark the episode as unmatched. Alternatively I might set the found episode as the "current", but this might cause the match to be a completely different show. In both case you need a Manuel step, what do you think?

rafaelgomesxyz:
Sorry I still haven't reviewed your PRs, time has been scarce lately. I have plans to do it later today.

package.json
rafaelgomesxyz: Isn't it better to just setup a `.gitattributes` file? I think using Linux endings is better, and Prettier offers a step-by-step on how to do that: https://prettier.io/docs/en/options.html#end-of-line

rafaelgomesxyz (dismissed):

package.json
rafaelgomesxyz: I've added a `.gitattributes` file to the master branch, so I don't think this is needed anymore.

MrMamen: I have a feeling that this is either already wrong, or it doesn't work, I get this on master even after your `.gitattributes` on windows:
```bash
$ npm run format-check

> universal-trakt-scrobbler@1.0.0 format-check 
> npm run prettier -- --check


> universal-trakt-scrobbler@1.0.0 prettier 
> prettier "**/*.{js,jsx,ts,tsx}" "--check"

Checking formatting...
[warn] src\api\TraktSearch.ts
[warn] src\common\BrowserStorage.tsx
[warn] src\common\Events.ts
[warn] src\common\Requests.ts
[warn] src\global.d.ts
[warn] src\modules\history\history.tsx
[warn] src\modules\options\components\OptionsListItem.tsx
[warn] src\modules\options\options.tsx
[warn] src\modules\options\OptionsApp.tsx
[warn] src\modules\popup\popup.tsx
[warn] src\modules\popup\PopupApp.tsx
[warn] Code style issues found in the above file(s). Forgot to run Prettier?
```

rafaelgomesxyz: You have to re-clone your local clone for it to work.

rafaelgomesxyz (approved):

PR #43, opened 2020-10-27
Implement various visual improvements

### LINKED ISSUE #34: Wrapped when logged in
Possible just on Norwegian localization, but the top bar is wrapped when being logged in:
loggedout
loggedin

rafaelgomesxyz:
Changes:

- [x] Migrate from CSS to Sass (not a visual change, but it makes implementing styles easier)
- [x] Make sure cards are aligned in sync page
- [x] Move sync options to sidebar
- [x] Use icons in popup (fixes #34)
- [x] Change the progress text to a progress bar in sync page
- [x] Use icons in history page
- [x] Use a CC background image


rafaelgomesxyz:
I think I'm done here.

webpack.config.ts
```diff
@@ -41,6 +41,9 @@ const loaders = {
 	css: {
 		loader: 'css-loader',
```
MrMamen, line 42: Is this still needed?

MrMamen (approved):
Looks great. I have gotten an warning with chrome lately:
https://developer.chrome.com/blog/enabling-shared-array-buffer/

Would you consider upgrading all dependencies in package.json? It could be a separate PR if you prefer.

webpack.config.ts
rafaelgomesxyz: Yes, because Sass compiles to CSS, so it still needs to load the CSS.

rafaelgomesxyz:
> Would you consider upgrading all dependencies in package.json? It could be a separate PR if you prefer.

I plan on making a PR for general cleaning, so I think it fits better there.

PR #49, opened 2020-11-05
Implement various pagination improvements

### LINKED ISSUE #45: Implement previous page

rafaelgomesxyz:
- [x] Fix bug that reloads history when reloading service page (without this fix, if you load a service page, then go to the home page and reload the same service page again, it will reload the history, resulting in duplicate items)
- [x] Disable next page button when there isn't a next page
- [x] Implement previous page (fixes #45)
- [x] Only load Trakt/TMDB/Corrections data for items in the current visible page (prevents unnecessary requests)
- [x] Only allow selecting/syncing items in the current visible page
- [x] Immediately update page to reflect changes to items per load option
- [x] Change data loading order (more useful to load correction suggestions before images)
- [x] Fix bug that loads next page when reloading service page
- [x] Use axios (makes canceling requests easier)
- [x] Cancel requests when loading a new page or exiting (prevents unnecessary requests)


rafaelgomesxyz:
Done here. Noticed some other stuff that need to be fixed, but I'll put them in other PRs. Only missing the NO translation.

MrMamen (approved):

PR #50, opened 2020-11-05
Fix loading of images

MrMamen:
Found and fixed the error.


rafaelgomesxyz (approved):
Good catch, I don't know I missed that.

PR #56, opened 2020-12-16
Set Viaplay region and use it for API calls

### LINKED ISSUE #55: Failing to retrieve more watched items on domains other than viaplay.no
In history sync for Viaplay, if you're using any other region than .no, it will fail to retrieve more items than the initial items. No matter if you increase the "Items per load" or try loading next page. 
Looking at the console I'm getting unauthorized calls to the watched-api on viaplay.no, so it's trying to request history from the Norwegian region, even though I'm not in that region.

smoothlystable:
Closes #55


MrMamen (dismissed):
Thanks for your contribution and for finding this

rafaelgomesxyz (approved):

PR #62, opened 2021-05-02
Inject content scripts dynamically

### LINKED ISSUE #51: Not able to disable streaming services with scrobble
The streaming services which support scrobbling is unable to deactivate once activated.

rafaelgomesxyz:
Fixes #51 

This was a tough one to fix, because `browser.tabs.onUpdated` is triggered multiple times for the same tab, and it's triggered even more by single-page apps like Netflix, so initially I couldn't figure out how to differentiate between a page load and a reload without resulting in duplicate injections.

Then I found out about `browser.webNavigation.onCommitted`, which allows detecting page loads / reloads and is only triggered once. However, the ability to detect page loads is not supported by Firefox, so I ended up having to use `browser.tabs.onUpdated` to detect page loads and `browser.webNavigation.onCommitted` to detect reloads.

A bit complex, but seems to be working well.

Will rebase once #49 is merged.

### DIFF
--- a/package.json
+++ b/package.json
@@ -51,7 +51,7 @@
 		"react-router-dom": "^5.2.0",
 		"rollbar": "^2.19.3",
 		"typeface-roboto": "^0.0.75",
-		"webextension-polyfill": "^0.6.0"
+		"webextension-polyfill": "^0.8.0"
 	},
 	"devDependencies": {
 		"@babel/core": "^7.11.4",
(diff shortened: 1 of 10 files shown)


rafaelgomesxyz:
I realized that the `tabs` permission isn't required, so I made it optional.

MrMamen (approved):

Issue #47, opened 2020-11-04
Give user option to do initial sync and prevent loading synced history


rafaelgomesxyz:
This would be very nice to prevent unnecessary requests to the Trakt API, specially since they're imposing limits now. Here's how I think it should happen:

1. Upon opening the sync page for a streaming service for the first time, the user is prompted with an option to sync their entire history so far (since the Trakt API allows 1000 requests every 5 minutes, it's perfect for scenarios like this)
2. We store the date of this initial sync (let's call it date A)
3. In the future, every time the user opens the sync page for that same streaming service again, we make sure that only history newer than date A is loaded
4. Additionally, every time we detect that the user has synced all history from date A to a date B, we update date A with date B

Not only would this save a ton of requests, but I believe it would also enhance the user experience, since they would only see items that they haven't synced. What do you think?

MrMamen:
That makes sense. Just have to make sure that it's possible to override or reset.

PR #63, opened 2021-05-03
Implement automatic sync

### LINKED ISSUE #8: Add option to sync automatically every x days

### LINKED ISSUE #11: Selected entries needs to be reset when syncing
If you sync the entries ar no longer selectable due to being found on trakt, they are however still selected, so syncing another time will send them to trakt as well.

rafaelgomesxyz:
Will fix #8 and fix #11 and lay down the foundation for #47

Still have some stuff to do, but wanted to discuss things.

I changed the UI for enabling/disabling services (as seen in the image below) to include an option to enable the automatic sync and decided to add separate options for scrobble/sync as well. If a switch is greyed out it's because the service doesn't support it. Let me know what you think.

New UI

How it works right now:

1. For the automatic sync to work, the user has to do a manual sync first, for the date to be saved (this will be good for #47).
2. After the user does the first manual sync, the background page of the extension will start checking, every time the user opens the browser, if it's been the number of days specified in the options since the last sync.
3. If it's time to sync again, the sync will happen silently in the background page. Only items with no `percentageWatched` (because that info is not available for some services) or items with `percentageWatched` higher or equal to 10 are synced (this number is arbitrary and can be changed).

One of the downsides is that all new items since the last sync will be synced, which might include items that the user hasn't actually finished watching yet (can be overcome with `percentageWatched`, but Netflix, for example, doesn't have that info).

Another downside is that you can't correct items, so if an item matched with a wrong item on Trakt, the wrong item will be synced, and I'm not entirely sure how to deal with this.

I'm currently working on these issues:

- [x] Indicate that the sync is happening somehow (maybe change the extension icon like it happens for scrobbles?)
- [x] Allow the user to visualize what was synced / what failed to sync somehow (maybe add a button to the extension popup that the user can click and be taken to a summary page of the last sync?)
- [x] Allow the user to specify a threshold for `percentageWatched`

Thoughts?

Will rebase once #62 is merged.

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,7 +42,7 @@ If you want to scrobble / sync from Netflix, this is the only Trakt.tv [plugin](
 ### Which streaming services are supported?
 
 - Amazon Prime (Scrobble only)
-- HBO Go (Scrobble only - tested only for Latin America)
+- HBO Go (Only tested for Latin America, can only sync the 30 last watched items)
 - Netflix
 - NRK
 - Viaplay (Sync only)
--- a/README.md
+++ b/README.md
@@ -42,7 +42,7 @@ If you want to scrobble / sync from Netflix, this is the only Trakt.tv [plugin](
 ### Which streaming services are supported?
 
 - Amazon Prime (Scrobble only)
-- HBO Go (Only tested for Latin America, can only sync the 30 last watched items)
+- HBO Go (Only available for Latin America, can only sync the 30 last watched items)
 - Netflix
 - NRK
 - Viaplay (Sync only)
--- a/src/_locales/nb/messages.json
+++ b/src/_locales/nb/messages.json
@@ -366,6 +366,9 @@
 	"wrongItemDialogInvalidUrlLabel": {
 		"message": "Lim inn en gyldig URL til Trakt!"
 	},
+	"wrongItemDialogSyncedWarning": {
+		"message": "WARNING: You are correcting an item that was already synced. If you proceed, the previously synced item will be removed from your Trakt history."
+	},
 	"yes": {
 		"message": "Ja"
 	}
--- a/src/_locales/nb/messages.json
+++ b/src/_locales/nb/messages.json
@@ -228,9 +228,6 @@
 	"no": {
 		"message": "Nei"
 	},
-	"noMoreHistory": {
-		"message": "Slutten av historikken er nådd."
-	},
 	"noStreamingServices": {
 		"message": "Ingen strømmetjenester er aktivert. Strømmetjenester kan aktiveres fra siden for alternativer."
 	},
@@ -262,6 +259,12 @@
 				"content": "$1"
 			}
 		}
+	},
+	"reachedHistoryEnd": {
+		"message": "Slutten av historikken er nådd."
+	},
+	"reachedLastSyncDate": {
+		"message": "Reached the last sync date. If you wish to load items past this point, clear the last sync date for this service in the options."
 	},
 	"readMore": {
 		"message": "Les mer"
--- a/src/_locales/nb/messages.json
+++ b/src/_locales/nb/messages.json
@@ -9,7 +9,7 @@
 		"message": "Legg til"
 	},
 	"addDate": {
-		"message": "Add date"
+		"message": "Legg til dato"
 	},
 	"addMissingWatchedDateFailed": {
 		"message": "Kunne ikke legge til manglende «sett dato»."
@@ -27,13 +27,13 @@
 		"message": "En universell sync-tjeneste for Trakt.tv."
 	},
 	"autoSync": {
-		"message": "Automatic Sync"
+		"message": "Synkroniser automatisk"
 	},
 	"autoSyncDescription": {
-		"message": "The number of days determines how often the automatic sync will happen for the streaming service. For the automatic sync to be activated, you must sync manually in the history page of the streaming service first, so that the last sync date can be saved.\n\nIt's also important to note that when loading the history of the streaming service, it will only load the history until the last item that was automatically synced. To reset this behavior, you have to clear the last sync date for the streaming service."
+		"message": "Antall dager avgjør hvor regelmessig automatisk synkronisering vil forekomme for den angitte strømmetjenesten. For å aktivere automatisk synkronisering må du først synkronisere manuelt i historikken en gang, slik at en synkroniseringsdato kan lagres.\n\nNB! Når du laster inn historikken til strømmetjenesten så vil den kun laste inn historikken opp til datoen på siste automatiske synkronisering. For å tilbakestille dette, må du tømme siste historikkdato for den gitte strømmetjenesten."
 	},
 	"autoSyncPageMessage": {
-		"message": "These are the items that were automatically synced or failed to automatically sync since the last review. If any corrections are needed, apply them, select the items that need to be corrected, and click the Sync button. When you do that, the cache will be deleted and these items will no longer appear on this page. If no corrections are needed, simply click the Clear Sync Cache button and the same thing will happen."
+		"message": "Dette er elementer som ble automatisk synkronisert eller mislykkes å synkronisere siden forrige gang. Dersom korreksjoner trengs, legg dem til, og velg elementene som trengs korreksjoner, og trykk på Synkroniser-knappen. Når du gjør det, slettes cachen og disse elementene vil ikke så på denne siden lengre. Dersom ingen korreksjoner er nødvendige, kan du trykke på Fjern synkroniseringscache-knappen og det samme vil skje."
 	},
 	"cancel": {
 		"message": "Avbryt"
@@ -45,7 +45,7 @@
 		"message": "Enkelte elementer mangler «sett dato». Legg til hver enkelt, eller tillat valget å legge til utgivelsesdato og forsøk igjen."
 	},
 	"clearLastSync": {
-		"message": "Clear Last Sync Date"
+		"message": "Fjern siste synkroniseringsdato"
 	},
 	"clearStorage": {
 		"message": "Fjern lagrede data"
@@ -57,10 +57,10 @@
 		"message": "Lagret data ble fjernet."
 	},
 	"clearSyncCache": {
-		"message": "Clear Sync Cache"
+		"message": "Fjern synkroniseringscache"
 	},
 	"clearTraktCache": {
-		"message": "Fjern Trakt cache"
+		"message": "Fjern Trakt-cache"
 	},
 	"clearTraktCacheFailed": {
 		"message": "Klarte ikke å fjerne Trakt cache."
@@ -73,10 +73,10 @@
 		"message": "Lukk"
 	},
 	"confirmClearLastSyncMessage": {
-		"message": "You'll need to sync manually in the history page of the streaming service again to reactivate the automatic sync."
+		"message": "Du er nødt til å synkronisere på nytt manuelt på historikksiden til strømmetjenesten om du ønsker å reaktivere automatisk synkronisering."
 	},
 	"confirmClearLastSyncTitle": {
-		"message": "Clear last sync date for $SERVICE$?",
+		"message": "Fjern siste synkroniseringsdato for $SERVICE$?",
 		"placeholders": {
 			"service": {
 				"content": "$1"
@@ -90,10 +90,10 @@
 		"message": "Fjern lagrede data?"
 	},
 	"confirmClearSyncCacheMessage": {
-		"message": "All of the items in the current page will be removed from the cache and you'll no longer be able to review them."
+		"message": "Alle elementene på siden blir fjernet fra cache, og du vil ikke lengre kunne se over dem."
 	},
 	"confirmClearSyncCacheTitle": {
-		"message": "Clear sync cache?"
+		"message": "Fjerne synkroniseringscachen?"
 	},
 	"confirmClearTraktCacheMessage": {
 		"message": "Alt rettelser du har gjort blir tilbakestilt."
@@ -118,7 +118,7 @@
 		"message": "Mørkt"
 	},
 	"days": {
-		"message": "Days"
+		"message": "Dager"
 	},
 	"errorNotification": {
 		"message": "Noe har gått galt. :("
@@ -195,7 +195,7 @@
 		"message": "Logg ut"
 	},
 	"minPercentageWatchedName": {
-		"message": "Minimum percentage watched."
+		"message": "Minimum prosent sett."
 	},
 	"missingWatchedDate": {
 		"message": "Mangler «sett dato»"
@@ -209,7 +209,7 @@
 		}
 	},
 	"missingWatchedDateDialogContentMultiple": {
-		"message": "Choose how you want to add the missing watched date for the $ITEMS$ items.",
+		"message": "Velg hvordan du ønsker å legge til de manglende sett-datoene for $ITEMS$ elementer.",
 		"placeholders": {
 			"items": {
 				"content": "$1"
@@ -267,10 +267,10 @@
 		"message": "Les mer"
 	},
 	"recentAutoSync": {
-		"message": "Some items were recently automatically synced. Click here to review them and make sure that the correct items were synced."
+		"message": "Enkelte elementer ble nylig automatisk synkronisert. Trykk her for å se på dem og pass på at de riktige elementene ble synkronisert."
 	},
 	"recentAutoSyncError": {
-		"message": "Some items recently failed to automatically sync. Click here to review and correct them."
+		"message": "Enkelte elementer ble ikke synkronisert automatisk. Trykk her for å rette dem."
 	},
 	"save": {
 		"message": "Lagre"
@@ -312,7 +312,7 @@
 		"message": "Send / motta forslag til rettelser."
 	},
 	"service": {
-		"message": "Service"
+		"message": "Strømmetjeneste"
 	},
 	"serviceScrobble": {
 		"message": "Scrobble"
@@ -396,7 +396,7 @@
 		"message": "Lim inn en gyldig URL til Trakt!"
 	},
 	"wrongItemDialogSyncedWarning": {
-		"message": "WARNING: You are correcting an item that was already synced. If you proceed, the previously synced item will be removed from your Trakt history."
+		"message": "ADVARSEL: Du retter nå et element som allerede har blitt synkronisert. Dersom du fortsetter blir den tidligere synkroniserte elementet fjernet fra Trakt-historikken din."
 	},
 	"yes": {
 		"message": "Ja"
--- a/src/_locales/nb/messages.json
+++ b/src/_locales/nb/messages.json
@@ -82,12 +82,6 @@
 	"darkTheme": {
 		"message": "Mørkt"
 	},
-	"disableScrobblingDescription": {
-		"message": "Dersom dette valget er aktivert vil utvidelsen ikke automatisk scrobble det du ser på. Dette er nyttig om du bare ønsker å bruke sync av historikk."
-	},
-	"disableScrobblingName": {
-		"message": "Deaktiver scrobbling."
-	},
 	"errorNotification": {
 		"message": "Noe har gått galt. :("
 	},
@@ -264,6 +258,15 @@
 	},
 	"sendReceiveSuggestionsName": {
 		"message": "Send / motta forslag til rettelser."
+	},
+	"service": {
+		"message": "Service"
+	},
+	"serviceScrobble": {
+		"message": "Scrobble"
+	},
+	"serviceSync": {
+		"message": "Sync"
 	},
 	"showNotificationsDescription": {
 		"message": "Viser nettleservarsler hver gang du starter / stopper å scrobble noe."
--- a/src/_locales/nb/messages.json
+++ b/src/_locales/nb/messages.json
@@ -8,6 +8,9 @@
 	"add": {
 		"message": "Legg til"
 	},
+	"addDate": {
+		"message": "Add date"
+	},
 	"addMissingWatchedDateFailed": {
 		"message": "Kunne ikke legge til manglende «sett dato»."
 	},
@@ -189,6 +192,14 @@
 		"message": "Hvordan ønsker du å legge til manglede «sett datoer» for $TITLE$?",
 		"placeholders": {
 			"title": {
+				"content": "$1"
+			}
+		}
+	},
+	"missingWatchedDateDialogContentMultiple": {
+		"message": "Choose how you want to add the missing watched date for the $ITEMS$ items.",
+		"placeholders": {
+			"items": {
 				"content": "$1"
 			}
 		}
--- a/src/_locales/nb/messages.json
+++ b/src/_locales/nb/messages.json
@@ -23,6 +23,12 @@
 	"appDescription": {
 		"message": "En universell sync-tjeneste for Trakt.tv."
 	},
+	"autoSync": {
+		"message": "Automatic Sync"
+	},
+	"autoSyncDescription": {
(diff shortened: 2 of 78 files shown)


rafaelgomesxyz:
Wait, nevermind, Netflix does offer percentage watched. Fixed it.

Also added an option to filter items by minimum percentage watched, so now the automatic sync picks up the value specified in that option.

rafaelgomesxyz:
Think I'm done here.

This is how it ended up working:

1. For the automatic sync to work, the user has to do a manual sync first, for the date to be saved (this will be good for #47).
2. After the user does the first manual sync, the background page of the extension will start checking, every hour, if it's been the number of days specified in the options since the last sync.
3. If it's time to sync again, the sync will happen silently in the background page (the extension icon will rotate to indicate that it's happening). Only items with no percentageWatched (because that info is not available for some services) or items with percentageWatched higher or equal to the value specified in the options are synced.
4. A button is added to the extension popup that opens a summary page where the user can see what was automatically synced / failed to sync. They can correct wrong items there.
5. If the user tries to load the history for the service that was automatically synced, it will stop loading the history once it reaches the last item that was synced (also good for #47). This can be reset in the options.

I think that's how it should work. Let me know if you think I need to change something.

rafaelgomesxyz:
~Wait, I think I found a couple bugs that I still need to fix:~

* ~Disabling the automatic sync seems to remove permissions.~
* ~If the user syncs the first history page, it won't load any more pages, because the date from the first page will be saved as the last sync date, and the next pages are older. A possible solution would be to store a copy of the last sync date and use that for as long as the user keeps loading pages and syncing. The only problem would be that if the user reloads/leaves the page, it will go back to not loading any more pages, but in that case the user could just go to the options page and reset the last sync date.~

MrMamen:
> Will fix #8 and lay down the foundation for #47
> 
> Still have some stuff to do, but wanted to discuss things.
> 
> I changed the UI for enabling/disabling services (as seen in the image below) to include an option to enable the automatic sync and decided to add separate options for scrobble/sync as well. If a switch is greyed out it's because the service doesn't support it. Let me know what you think.

Hmm. I think it's better with a text, "not supported", or an icon with the same as a tooltip.
 
> How it works right now:
> 
> 1. For the automatic sync to work, the user has to do a manual sync first, for the date to be saved (this will be good for #47).
> 2. After the user does the first manual sync, the background page of the extension will start checking, every time the user opens the browser, if it's been the number of days specified in the options since the last sync.
> 3. If it's time to sync again, the sync will happen silently in the background page. Only items with no `percentageWatched` (because that info is not available for some services) or items with `percentageWatched` higher or equal to 10 are synced (this number is arbitrary and can be changed).

I think 80 or 90 should be the default value here.
 
> One of the downsides is that all new items since the last sync will be synced, which might include items that the user hasn't actually finished watching yet (can be overcome with `percentageWatched`, but Netflix, for example, doesn't have that info).
> 
> Another downside is that you can't correct items, so if an item matched with a wrong item on Trakt, the wrong item will be synced, and I'm not entirely sure how to deal with this.

Hmm. That't tricky. Also some items doesn't match anything, but they can be manually fixed later when syncing the old way, or will the date prevent this from be fixed?

> I'm currently working on these issues:
> 
> * [x]  Indicate that the sync is happening somehow (maybe change the extension icon like it happens for scrobbles?)
> * [x]  Allow the user to visualize what was synced / what failed to sync somehow (maybe add a button to the extension popup that the user can click and be taken to a summary page of the last sync?)
> * [x]  Allow the user to specify a threshold for `percentageWatched`
> 
> Thoughts?
You could open a page when a sync has occurred with info, but this might be annoying. So perhaps your idea is better.

README.md
```diff
@@ -42,7 +42,7 @@ If you want to scrobble / sync from Netflix, this is the only Trakt.tv [plugin](
 ### Which streaming services are supported?
 
 - Amazon Prime (Scrobble only)
-- HBO Go (Scrobble only - tested only for Latin America)
+- HBO Go (Only tested for Latin America, can only sync the 30 last watched items)
```
MrMamen: I think it would be more fair to say that only Latin America is currently supported. I mean the URL is hardcoded to hbogola is it not?

src/_locales/nb/messages.json
```diff
@@ -8,6 +8,9 @@
 	"add": {
 		"message": "Legg til"
 	},
+	"addDate": {
+		"message": "Add date"
```
MrMamen, line 12: Legg til dato

src/_locales/nb/messages.json
```diff
@@ -23,6 +26,15 @@
 	"appDescription": {
 		"message": "En universell sync-tjeneste for Trakt.tv."
 	},
+	"autoSync": {
+		"message": "Automatic Sync"
```
MrMamen, line 30: Synkroniser automatisk

src/_locales/nb/messages.json
```diff
@@ -23,6 +26,15 @@
...
+	},
+	"autoSyncDescription": {
+		"message": "The number of days determines how often the automatic sync will happen for the streaming service. For the automatic sync to be activated, you must sync manually in the history page of the streaming service first, so that the last sync date can be saved.\n\nIt's also important to note that when loading the history of the streaming service, it will only load the history until the last item that was automatically synced. To reset this behavior, you have to clear the last sync date for the streaming service."
```
MrMamen, line 33: Antall dager avgjør hvor regelmessig automatisk synkronisering vil forekomme for den angitte strømmetjenesten. For å aktivere automatisk synkronisering må du først synkronisere manuelt i historikken en gang, slik at en synkroniseringsdato kan lagres.\n\nNB! Når du laster inn historikken til strømmetjenesten så vil den kun laste inn historikken opp til datoen på siste automatiske synkronisering.

src/_locales/nb/messages.json
```diff
@@ -23,6 +26,15 @@
...
+	},
+	"autoSyncPageMessage": {
+		"message": "These are the items that were automatically synced or failed to automatically sync since the last review. If any corrections are needed, apply them, select the items that need to be corrected, and click the Sync button. When you do that, the cache will be deleted and these items will no longer appear on this page. If no corrections are needed, simply click the Clear Sync Cache button and the same thing will happen."
```
MrMamen, line 36: Dette er elementer som ble automatisk synkronisert eller mislykkes å synkronisere siden forrige gang.  Dersom korreksjoner trengs, legg dem til, og velg elementene som trengs korreksjoner, og trykk på Synkroniser-knappen. Når du gjør det, slettes cachen og disse elementene vil ikke så på denne siden lengre. Dersom ingen korreksjoner er nødvendige, kan du trykke på Fjern synkroniseringscache-knappen og det samme vil skje.

src/_locales/nb/messages.json
```diff
@@ -32,6 +44,9 @@
 	"cannotSyncMissingWatchedDate": {
 		"message": "Enkelte elementer mangler «sett dato». Legg til hver enkelt, eller tillat valget å legge til utgivelsesdato og forsøk igjen."
 	},
+	"clearLastSync": {
+		"message": "Clear Last Sync Date"
```
MrMamen, line 48: Fjern siste synkroniseringsdato

src/_locales/nb/messages.json
```diff
@@ -41,6 +56,9 @@
 	"clearStorageSuccess": {
 		"message": "Lagret data ble fjernet."
 	},
+	"clearSyncCache": {
+		"message": "Clear Sync Cache"
```
MrMamen, line 60: Fjern synkroniseringscache

src/_locales/nb/messages.json
```diff
@@ -54,12 +72,29 @@
 		"description": "This is the verb, as in 'to close'.",
 		"message": "Lukk"
 	},
+	"confirmClearLastSyncMessage": {
+		"message": "You'll need to sync manually in the history page of the streaming service again to reactivate the automatic sync."
```
MrMamen, line 76: Du er nødt til å synkronisere på nytt manuelt på historikksiden til strømmetjenesten om du ønsker å reaktivere automatisk synkronisering.

src/_locales/nb/messages.json
```diff
@@ -54,12 +72,29 @@
...
+	},
+	"confirmClearLastSyncTitle": {
+		"message": "Clear last sync date for $SERVICE$?",
```
MrMamen, line 79: Fjern siste synkroniseringsdato for $SERVICE$?

src/_locales/nb/messages.json
```diff
@@ -54,12 +72,29 @@
...
+		}
+	},
 	"confirmClearStorageMessage": {
 		"message": "Alle valgene vil bli tilbakestilt til standardvalgene og du blir nødt til å logge inn på nytt."
 	},
 	"confirmClearStorageTitle": {
 		"message": "Fjern lagrede data?"
 	},
+	"confirmClearSyncCacheMessage": {
+		"message": "All of the items in the current page will be removed from the cache and you'll no longer be able to review them."
```
MrMamen, line 93: Alle elementene på siden blir fjernet fra cache, og du vil ikke lengre kunne se over dem.

src/_locales/nb/messages.json
```diff
@@ -54,12 +72,29 @@
...
+	},
+	"confirmClearSyncCacheTitle": {
+		"message": "Clear sync cache?"
```
MrMamen, line 96: Fjerne synkroniseringscachen?

src/_locales/nb/messages.json
```diff
@@ -82,11 +117,8 @@
 	"darkTheme": {
 		"message": "Mørkt"
 	},
-	"disableScrobblingDescription": {
-		"message": "Dersom dette valget er aktivert vil utvidelsen ikke automatisk scrobble det du ser på. Dette er nyttig om du bare ønsker å bruke sync av historikk."
-	},
-	"disableScrobblingName": {
-		"message": "Deaktiver scrobbling."
+	"days": {
+		"message": "Days"
```
MrMamen, line 121: Dager

src/_locales/nb/messages.json
```diff
@@ -82,11 +117,8 @@
 	"darkTheme": {
 		"message": "Mørkt"
 	},
-	"disableScrobblingDescription": {
-		"message": "Dersom dette valget er aktivert vil utvidelsen ikke automatisk scrobble det du ser på. Dette er nyttig om du bare ønsker å bruke sync av historikk."
```
MrMamen, line 86: Do you remote the ability to disable scrobbling?

src/_locales/nb/messages.json
```diff
@@ -162,6 +194,9 @@
 	"logout": {
 		"message": "Logg ut"
 	},
+	"minPercentageWatchedName": {
+		"message": "Minimum percentage watched."
```
MrMamen, line 198: Minimum prosent sett.

src/models/TraktItem.ts
```diff
@@ -1,4 +1,4 @@
-import { Moment } from 'moment';
+import * as moment from 'moment';
```
MrMamen, line 1: Why this change?

src/models/TraktItem.ts
```diff
@@ -61,4 +70,22 @@ export class TraktItem implements ITraktItem {
...
+			watchedAt: item.watchedAt?.unix(),
+			progress: item.progress,
+		};
+	};
+
+	static load = (savedItem: SavedTraktItem): TraktItem => {
+		const options: ITraktItem = {
+			...savedItem,
+			watchedAt:
+				typeof savedItem.watchedAt !== 'undefined' ? moment(savedItem.watchedAt * 1e3) : undefined,
```
MrMamen, line 82: What does ie3 mean?

MrMamen (commented):
Wow. Huge amount of work here. A bit hard to review all the code, but I will checkout and test this.

src/_locales/nb/messages.json
rafaelgomesxyz: Since you can now enable scrobbling separately for each streaming service, I don't see the point for a global option. You think there should be one?

src/models/TraktItem.ts
rafaelgomesxyz: Because I'm now converting a date in the `load` method.

src/models/TraktItem.ts
rafaelgomesxyz: 10^3 = 1000, basically just converting from seconds to milliseconds.

rafaelgomesxyz:
> Wow. Huge amount of work here. A bit hard to review all the code, but I will checkout and test this.

Yeah, sorry about that. I didn't know if I should make smaller PRs, because it feels a bit spammy. Going commit-by-commit should be easier than going by files changed, because I tried to keep changes where they belong in each commit.

src/modules/options/components/StreamingServiceOptions.tsx
```diff
@@ -1,23 +1,60 @@
...
-			{(Object.entries(options) as [StreamingServiceId, boolean][])
-				.sort(([idA], [idB]) => idA.localeCompare(idB))
-				.map(([id, value]) => (
-					<StreamingServiceOption key={id} id={id} value={value} />
-				))}
-		</List>
+		<ListItem>
+			<Grid container>
+				<Grid container className="options-grid-container">
+					<Grid item className="options-grid-item options-grid-item--centered" xs={4}>
```
MrMamen: I don't think it looks to good when this is centered.

MrMamen (changes requested):
I seem to get this error when trying to load history (manually) on all pages:
`Access to XMLHttpRequest at 'https://www.netflix.com/Activate' from origin 'chrome-extension://bgdoldpgfpjehilgjeophjhlpemiopbi' has been blocked by CORS policy: No 'Access-Control-Allow-Origin' header is present on the requested resource.`
I get this also on master, so not anything wrong with this PR, but just need to investigate if it's a new chrome-verison or something else. Also happens on other streaming services.

src/models/TraktItem.ts
MrMamen: Right. Strange notation. In my mind it would be 10e3, but I see that you are right.

MrMamen:
> > Wow. Huge amount of work here. A bit hard to review all the code, but I will checkout and test this.
> 
> Yeah, sorry about that. I didn't know if I should make smaller PRs, because it feels a bit spammy. Going commit-by-commit should be easier than going by files changed, because I tried to keep changes where they belong in each commit.

No need to apologize. I'm not always able to review things ASAP, so if having things together makes more sense to you, just to it that way. I'm fine with either.

rafaelgomesxyz:
> I seem to get this error when trying to load history (manually) on all pages:
> `Access to XMLHttpRequest at 'https://www.netflix.com/Activate' from origin 'chrome-extension://bgdoldpgfpjehilgjeophjhlpemiopbi' has been blocked by CORS policy: No 'Access-Control-Allow-Origin' header is present on the requested resource.`
> I get this also on master, so not anything wrong with this PR, but just need to investigate if it's a new chrome-verison or something else. Also happens on other streaming services.

Probably a permission error? Did you try disabling and enabling the Netflix sync again?

MrMamen:
> Probably a permission error? Did you try disabling and enabling the Netflix sync again?

I think there are some migrating problems with BrowserStorage.
When using the 'clearStorage' on master I get it to work again. When switching to this branch I've been experienced a few strange issues.  Like:
`TypeError: Cannot read property 'amazon-prime' of undefined
    at _BrowserStorage.addOption (background.js:113956)`
  
When using  'clearStorage'  on this branch I was not even able to log in again. I think the browser storage should be more stable as switching to branched with different set of options shouldn't break the extension. Instead extra options should be silently ignored, or logged.

src/modules/options/components/StreamingServiceOptions.tsx
rafaelgomesxyz: Just the service one?

MrMamen: Yeah. It seems to stand out as it has some distance to the text below.

rafaelgomesxyz:
Better? The ban icon shows "Not Available" when you hover over it.

rafaelgomesxyz:
Did you forget to translate `serviceSync` or is Sync just Sync in NB? Also, I'd wait to translate until I fix the other stuff because I'm adding more messages.

README.md
rafaelgomesxyz: True, I actually think HBO Go is only available for Latin America. On US, it's HBO Max, and I'm not sure about the rest of the world.

MrMamen:
Let me know when translations are ready for another overhaul. When testing this I was not able to get the auto sync to work. I've put the value to 1 day, but I'm not quite sure why it's not working as expected.

rafaelgomesxyz:
> Let me know when translations are ready for another overhaul. When testing this I was not able to get the auto sync to work. I've put the value to 1 day, but I'm not quite sure why it's not working as expected.

Did you sync manually first and wait 24 hours? You can trigger the automatic sync sooner by commenting this line: https://github.com/trakt-tools/universal-trakt-scrobbler/pull/63/files#diff-6504efd683d6af1367f550ccb51266b4c3989bf158e0d4091225a626698c315dR209

rafaelgomesxyz:
Made some improvements:

* **Added a button to continue loading history past the last sync date:** this prevents the user from having to go to the options and clear the last sync date if they want to load older history, and makes it more straightforward
* **Replaced browser history with hash history:** this makes it possible not only to refresh pages, but also to go directly to specific pages (for example, the user can go directly to the Netflix history page without having to select it in the service list first)
* **Improved options layout:** the external link icon allows the user to go directly to the history page of the service, and the error icon appears when the user hasn't synced manually yet, to indicate that the automatic sync is not activated (I think this makes it easier to understand how it works?)

src/modules/background/background.ts
```diff
@@ -150,34 +166,63 @@ export interface NavigationCommittedParams {
...
+		StreamingServiceValue
+	][];
+	const scrobblerEnabled = streamingServiceEntries.some(
+		([streamingServiceId, value]) =>
+			streamingServices[streamingServiceId].hasScrobbler && value.scrobble
+	);
+	if (scrobblerEnabled) {
+		addWebNavigationListener(BrowserStorage.options);
+
+		void checkServicesToSync();
```
MrMamen: I don't see why scrobbling should be a requirement for auto-sync

MrMamen (changes requested):
Also I get this error in the console, but it seems like everything is working:
Error in event handler: TypeError: Cannot set property 'value' of undefined
    at _BrowserStorage.addOption

src/modules/background/background.ts
rafaelgomesxyz: Oh that's definitely in the wrong place, good catch!

rafaelgomesxyz:
> Also I get this error in the console, but it seems like everything is working:
> Error in event handler: TypeError: Cannot set property 'value' of undefined
> at _BrowserStorage.addOption

Noted, I'll see if I can reproduce and fix. I have a few more commits to push here and then I should be done.

MrMamen:
Great. Let me know when it's time for translations.

rafaelgomesxyz:
Main improvements from the most recent commits:

* **Made it show the last sync date when hovering over the Clear Last Sync button:** allows the user to quickly see when the last sync happened
* **Changed how adding with release date works:** now when the option to add with release date is enabled, the release date will be shown in the item cards and the history for the items will be retrieved again, to see if they've already been synced with release date; when it's disabled, the original watched date (if there was any) will be shown in the item cards
* **Added option to add with release date only for items that are missing the watched date:** this makes it easier to have the option enabled, but only for items that are missing the watched date, leaving items that have the watched date intact

I think you can update the translations now.

> Also I get this error in the console, but it seems like everything is working:
Error in event handler: TypeError: Cannot set property 'value' of undefined
at _BrowserStorage.addOption

I was not able to reproduce this. Where does it happen?

src/common/BrowserStorage.tsx
```diff
@@ -248,26 +319,85 @@ class _BrowserStorage {
...
+						...value,
+					};
+					this.optionsDetails[option.id].value[id] = {
+						...this.optionsDetails[option.id].value[id],
+						...value,
+					};
+				}
+			} else {
+				this.options[option.id] = option.value;
+				this.optionsDetails[option.id].value = option.value;
```
MrMamen, line 536: I've pinpointed the previously mentioned error to this line, when option.id = "disableScrobbling". The optionDetails-lookup object is then undefined. The extension seems to remember enabling/disabling though.

MrMamen (commented):
Not quite sure how I managed to get into this state. But the sync-cache seems to duplicate. When going to the reivew-page I now get 22 duplicates of the same three items.

src/common/BrowserStorage.tsx
rafaelgomesxyz: Ah, I see, it's because `disableScrobbling` was removed. I just pushed a commit that upgrades/downgrades the storage if necessary, to avoid conflicts like this.

rafaelgomesxyz:
> Not quite sure how I managed to get into this state. But the sync-cache seems to duplicate. When going to the reivew-page I now get 22 duplicates of the same three items.

I guess it could be possible to duplicate items by clearing the last sync date and syncing an older item, because then the auto sync will go through the same items that it did before and add them to the cache. I should probably also clear the cache when the last sync date is cleared.

If you didn't do that, it could be failing to detect new items and always running through the same ones. Is this the NRK auto sync? One thing you could do is open the background page console when the auto sync is happening (you can always trigger it manually by commenting the `now - value.lastSync >= value.autoSyncDays * 86400` line) and take a look at the requests, to see if it's going through the same items over and over again.

src/modules/history/pages/SyncPage.tsx
```diff
@@ -0,0 +1,475 @@
...
+			});
+		}
+		setContent((prevContent) => ({
+			...prevContent,
+			isLoading: true,
+		}));
+		await TraktSync.sync(store.data.selectedItems);
+		if (serviceId) {
+			const lastSync =
+				store.data.selectedItems[0].watchedAt?.unix() ?? Math.trunc(Date.now() / 1e3);
```
MrMamen: While trying to recreate the other bug, I encountered another here. After clearing lastsync date I was not able to re-enable the autosync because lastSync date where never set.
At this point the `store.data.selectedItems` is an empty array, so it fails with `TypeError: Cannot read property 'watchedAt' of undefined`. The array is however not empty on the await statement on line 134 just above. My theory is that the sync updated the DOM, and thus resets the selectedItems. I think this bug is solved by creating a copy of the array before calling `TraktSync.sync(items)`

rafaelgomesxyz: Fixed

src/streaming-services/common/AutoSync.ts
```diff
@@ -59,7 +59,7 @@ class _AutoSync {
 					...BrowserStorage.options.streamingServices,
 					[serviceId]: {
 						...BrowserStorage.options.streamingServices[serviceId],
-						lastSync: items[0]?.watchedAt?.unix() ?? now,
+						lastSync: now,
```
rafaelgomesxyz, line 63: If we save the item watched date here, it could keep triggering the auto sync check every hour, so it's better to save the current date.

rafaelgomesxyz:
I can reproduce the duplicating cache issue now, trying to find the problem.

rafaelgomesxyz:
> I can reproduce the duplicating cache issue now, trying to find the problem.

Ah, I think I figured it out. It's probably because `SyncStore` is not being reset during the auto sync, so the items remain there, and if it auto syncs again in the same browser session, it'll add them to the cache again.

Just pushed the fix.

MrMamen (dismissed):
Good job! I noticed an annoyance. Since you can go from the options page to the sync page, in a new tab: The old Option Page still keeps the explanation mark, and you have to do a refresh to make it go away. Not a show-stopper, but a minor issue I guess. One option would be to not open in a new tab when clicking on the sync-icon.

rafaelgomesxyz:
I noticed that as well. I can just add a listener to the storage and refresh the component when the storage changes.

rafaelgomesxyz:
The component should refresh now. I think that's it, all done here.

MrMamen (approved):
Very nice work!

PR #65, opened 2021-05-21
Adding Disney Plus service

ITSYV:


src/streaming-services/disneyplus/DisneyplusParser.ts
```diff
@@ -0,0 +1,111 @@
...
+}
+
+class _DisneyplusParser implements ScrobbleParser {
+	id: string;
+	title: string;
+	episodeTitle: string;
+	type: string;
+	season: number;
+	episode: number;
+	isCollection: boolean;
```
rafaelgomesxyz, line 20: Any particular reason for having `title`, `episodeTitle`, `type`, `season`, `episode` and `isCollection` here? It doesn't look like they're used anywhere?

rafaelgomesxyz (changes requested):
Please run the linter and fix the errors (`npm run fix`).

MrMamen:
Also remember to update the readme file with the new service.

src/streaming-services/disneyplus/DisneyplusParser.ts
ITSYV: Disney Plus removes the elements from the DOM when hiding the title and controls.
If I remove them, it seems that the plugin is querying Trakt with unkown data.

I'm new to typescript, so not sure if this can be handled another way :)

ITSYV:
> Please run the linter and fix the errors (`npm run fix`).

src/streaming-services/disneyplus/DisneyplusParser.ts:62:36 - error TS2339: Property 'style' does not exist on type 'Element'.
It does exist though.
As said earlier, new to typescript and not sure how to handle this one.
It seems that subitems from querySelectorAll don't handle the "style" option, though querySelector does have it (and returns the same object)


Any idea how to fix / ignore this?

rafaelgomesxyz:
> > Please run the linter and fix the errors (`npm run fix`).
> 
> src/streaming-services/disneyplus/DisneyplusParser.ts:62:36 - error TS2339: Property 'style' does not exist on type 'Element'.
> It does exist though.
> As said earlier, new to typescript and not sure how to handle this one.
> It seems that subitems from querySelectorAll don't handle the "style" option, though querySelector does have it (and returns the same object)
> 
> Any idea how to fix / ignore this?

`querySelectorAll` returns `NodeListOf<Element>`, and `style` is only present in `HTMLElement`, so all you'd have to do is cast the type: `document.querySelectorAll(...) as NodeListOf<HTMLElement>` It's an annoying error, but unfortunately this is the only way to deal with it.

src/streaming-services/disneyplus/DisneyplusParser.ts
rafaelgomesxyz: But you're not using these specific properties anywhere, so I don't see how it can affect anything. You only assign them [here](https://github.com/trakt-tools/universal-trakt-scrobbler/pull/65/files#diff-d1fd825e3b0b628976a42f935a29148422a5aaa028b171432183015a39da4ddaR91-R96). If you remove those lines, does it still query Trakt with unknown data?

PR #66, opened 2021-05-21
Adding Disney Plus Service

ITSYV:


rafaelgomesxyz:
The tests are still failing. You ran `npm run fix` and it gave no errors?

Also, you don't have to create a new PR every time. Just push to your branch and the commits will appear here.

ITSYV:
> The tests are still failing. You ran `npm run fix` and it gave no errors?
> 
> Also, you don't have to create a new PR every time. Just push to your branch and the commits will appear here.

I only get errors on files I did not change

`scripts/generateRelease.js:45:5 - error TS2322: Type 'Buffer' is not assignable to type 'string'.

45     data: file.content,
       ~~~~

src/api/TraktAuth.ts:69:65 - error TS2345: Argument of type '(value: TraktAuthDetails | PromiseLike<TraktAuthDetails>) => void' is not assignable to parameter of type 'PromiseResolve<TraktAuthDetails>'.
  Types of parameters 'value' and 'value' are incompatible.
    Type 'TraktAuthDetails | PromiseLike<TraktAuthDetails> | undefined' is not assignable to type 'TraktAuthDetails | PromiseLike<TraktAuthDetails>'.
      Type 'undefined' is not assignable to type 'TraktAuthDetails | PromiseLike<TraktAuthDetails>'.

69    promise = new Promise((resolve) => void this.startManualAuth(resolve));
                                                                   ~~~~~~~

src/api/TraktAuth.ts:83:62 - error TS2345: Argument of type '(value: TraktAuthDetails | PromiseLike<TraktAuthDetails>) => void' is not assignable to parameter of type 'PromiseResolve<TraktAuthDetails>'.

83    return new Promise((resolve) => void this.startManualAuth(resolve));
                                                                ~~~~~~~

src/streaming-services/common/ScrobbleEvents.ts:37:11 - error TS1243: 'async' modifier cannot be used with 'abstract' modifier.

37  abstract async checkForChanges(): Promise<void>;`

ITSYV:
Code cleanup with the github errors checked :)
got no other errors with "npm run fix" except the 4 mentioned above

Only error I didn't fix was something with no await for an async method (parseProgress), but not sure how to fix that. It's a straight copy from the Netflix / Amazon files.

rafaelgomesxyz:
Why did you add the file `src/streaming-services.ts`?

ITSYV:
> Why did you add the file `src/streaming-services.ts`?

I got a message that 'disneyplus' should be disneyplus without the quotes, so I changed it.
And afterwards I noticed that I forgot to remove the streamz-be mention.

Building another service for streamz.be :)

ITSYV:
> > Why did you add the file `src/streaming-services.ts`?
> 
> I got a message that 'disneyplus' should be disneyplus without the quotes, so I changed it.
> And afterwards I noticed that I forgot to remove the streamz-be mention.
> 
> Building another service for streamz.be :)

ooh, now I see what you mean. 
I uploaded it here on github. not on my dev pc right now

rafaelgomesxyz:
I think most of the errors are because you're probably not running Prettier? Prettier should format all files automatically when you run `npm run fix`. Not sure why that's not happening for you.

rafaelgomesxyz:
For the `parseItem` error, you just have to remove `async` from the function if it doesn't have `await`.

ITSYV:
> I think most of the errors are because you're probably not running Prettier? Prettier should format all files automatically when you run `npm run fix`. Not sure why that's not happening for you.

reinstalled prettier, but still don't get other errors locally. Only the 4 above

There is no "async" on the parseProgress function..

rafaelgomesxyz:
> > I think most of the errors are because you're probably not running Prettier? Prettier should format all files automatically when you run `npm run fix`. Not sure why that's not happening for you.
> 
> reinstalled prettier, but still don't get other errors locally. Only the 4 above
> 
> There is no "async" on the parseProgress function..

I think it's in `parseItem`.

ITSYV:
Oh yep!
But isn't it supposed to be there? Also copied that from the other files.
I really have no clue about how that works :) I was already happy that it was working as expected 😄

rafaelgomesxyz:
> Oh yep!
> But isn't it supposed to be there? Also copied that from the other files.
> I really have no clue about how that works :) I was already happy that it was working as expected 😄

It's probably because in the other files `await` is used somewhere in the function, and you didn't need to use it.

If `await` isn't used in the function, then `async` shouldn't be there. `async` is only required in asynchronous functions.

ITSYV:
It seems that if I remove it, I get another error because it's not assignable to the same property in ScrobbleParser.

rafaelgomesxyz:
> It seems that if I remove it, I get another error because it's not assignable to the same property in ScrobbleParser.

I see. Can you try changing the line https://github.com/trakt-tools/universal-trakt-scrobbler/blob/master/src/streaming-services/common/ScrobbleController.ts#L12 to `parseItem(): Promise<Item | undefined> | Item | undefined;`? That should make it assignable.

ITSYV:
Done, next errors:
- TS2322: Type 'undefined' is not assignable to type 'Promise '.
- TS2739: Type 'Item' is missing the following properties from type 'Promise ': then, catch, [Symbol.toStringTag], finally

ITSYV:
Or should I change the parseItem function to: "parseItem = (): Item | undefined => {" in my parser file?

rafaelgomesxyz:
> Or should I change the parseItem function to: "parseItem = (): Item | undefined => {" in my parser file?

Yep, forgot to mention that.

rafaelgomesxyz:
Also forgot to mention that you need to remove `await` from `const item = await DisneyplusParser.parseItem();`, since it's no longer async.

Sorry for the hassle with the errors! This hasn't been a problem before.

ITSYV:
Updated!
No problem, if I knew more of Typescript, I probably would have noticed all these kinds of things. But never too old to learn I guess. 

If the Disney Plus addon is fully working and without errors, I'll give it a copy to add Streamz and some more Belgian streaming sites

rafaelgomesxyz:
Looks good! Thanks for contributing!

rafaelgomesxyz (approved):

rafaelgomesxyz:
Also, not sure if you're aware, but creating a separate branch to work on your PRs is better than working directly on master.

ITSYV:
I am now! But I was already making a branch for Streamz. Noticed that it would indeed be more convenient that way.

PR #70, opened 2021-05-23
Adding VRTNu - Belgian streaming service

ITSYV:
Another Belgian streaming service VRT Nu
Final one ;-)

### DIFF
--- a/src/streaming-services/vrtnu-be/VrtnuBeApi.ts
+++ b/src/streaming-services/vrtnu-be/VrtnuBeApi.ts
@@ -1,0 +1,24 @@
+import { Api } from '../common/Api';
+import { registerApi } from '../common/common';
+
+// Define any types you need here.
+
+// This class should communicate with the service API, in order to retrieve the necessary information for scrobbling. If the service does not have an API, that information should be retrieved in the *Parser class instead, so that this class only deals with requests, and not direct DOM manipulation. Keep in mind that some services might have hidden APIs that you can use (you can usually find them by watching your network requests when using the service).
+class _VrtnuBeApi extends Api {
+	// Define any properties you need here.
+
+	constructor() {
+		super('vrtnu-be');
+	}
+
+	// This method is only required for syncing, but since it is an abstract method, we have to implement at least a basic block for it.
+	loadHistory = (itemsToLoad: number): Promise<void> => {
+		return Promise.resolve();
+	};
+
+	// Define any methods you need here.
+}
+
+export const VrtnuBeApi = new _VrtnuBeApi();
+
+registerApi('vrtnu-be', VrtnuBeApi);
--- a/src/streaming-services/vrtnu-be/VrtnuBeApi.ts
+++ b/src/streaming-services/vrtnu-be/VrtnuBeApi.ts
@@ -1,12 +1,7 @@
 import { Api } from '../common/Api';
 import { registerApi } from '../common/common';
 
-// Define any types you need here.
-
-// This class should communicate with the service API, in order to retrieve the necessary information for scrobbling. If the service does not have an API, that information should be retrieved in the *Parser class instead, so that this class only deals with requests, and not direct DOM manipulation. Keep in mind that some services might have hidden APIs that you can use (you can usually find them by watching your network requests when using the service).
 class _VrtnuBeApi extends Api {
-	// Define any properties you need here.
-
 	constructor() {
 		super('vrtnu-be');
 	}
@@ -15,8 +10,6 @@ class _VrtnuBeApi extends Api {
 	loadHistory = (itemsToLoad: number): Promise<void> => {
 		return Promise.resolve();
 	};
-
-	// Define any methods you need here.
 }
 
 export const VrtnuBeApi = new _VrtnuBeApi();
--- a/src/streaming-services/vrtnu-be/VrtnuBeParser.ts
+++ b/src/streaming-services/vrtnu-be/VrtnuBeParser.ts
@@ -1,0 +1,74 @@
+import { Item } from '../../models/Item';
+import { registerScrobbleParser } from '../common/common';
+import { ScrobbleParser } from '../common/ScrobbleController';
+import { VrtnuBeApi } from './VrtnuBeApi';
+
+class _VrtnuBeParser implements ScrobbleParser {
+	id: string;
+	videoId: string;
+	progress: number;
+	isPaused: boolean;
+
+	constructor() {
+		this.id = '';
+		this.videoId = '';
+		this.progress = 0.0;
+		this.isPaused = false;
+	}
+
+	parseProgress = (): number => {
+		let progress = 0.0;
+		const scrubber: HTMLElement | null = document.querySelector('.vjs-play-progress');
+
+		if (scrubber) {
+			progress = parseFloat(scrubber?.style.left);
+			this.progress = progress;
+		}
+
+		return progress;
+	};
+
+	parseItem = (): Item | undefined => {
+		const serviceId = 'vrtnu-be';
+		let showTitle: string | null = null;
+		let seasonOrYear: string | null = null;
+		let subTitle: string | undefined = undefined;
+		const matches = /\/a-z\/(.+)\/(.+)\/(.+)\//.exec(location.href);
+
+		if (matches) {
+			[, showTitle, seasonOrYear, subTitle] = matches;
+		}
+
+		const id = subTitle ?? '';
+		const title = showTitle ?? '';
+		const episodeTitle = '';
+		const episodeId = subTitle?.substring(subTitle?.lastIndexOf('s' + seasonOrYear + 'a')) ?? '';
+		const type = episodeId ? 'show' : 'movie';
+		const season = type == 'show' ? parseInt(seasonOrYear ?? '') || 0 : 0;
+		const episode = type == 'show' ? parseInt(episodeId ?? '') || 0 : 0;
+		const year = type == 'movie' ? parseInt(seasonOrYear ?? '') : 0;
+		const isCollection = false;
+
+		if (showTitle) {
+			this.videoId = id;
+		} else {
+			return undefined;
+		}
+
+		return new Item({
+			serviceId,
+			id,
+			type,
+			title,
+			year,
+			episodeTitle,
+			season,
+			episode,
+			isCollection,
+		});
+	};
+}
+
+export const VrtnuBeParser = new _VrtnuBeParser();
+
+registerScrobbleParser('vrtnu-be', VrtnuBeParser);
--- a/src/streaming-services/vrtnu-be/VrtnuBeParser.ts
+++ b/src/streaming-services/vrtnu-be/VrtnuBeParser.ts
@@ -33,26 +33,28 @@ class _VrtnuBeParser implements ScrobbleParser {
 		let showTitle: string | null = null;
 		let seasonOrYear: string | null = null;
 		let subTitle: string | undefined = undefined;
-		const matches = /\/a-z\/(.+)\/(.+)\/(.+)\//.exec(location.href);
+		let seasonAndEpisode: string | undefined = undefined;
+		let id: string | undefined = undefined;
+		let seasonStr: string | undefined = undefined;
+		let episodeStr: string | undefined = undefined;
+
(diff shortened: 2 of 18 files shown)


src/streaming-services/vrtnu-be/VrtnuBeParser.ts
```diff
@@ -0,0 +1,77 @@
...
+		const matches = /\/a-z\/(.+)\/(.+)\/(.+)\//.exec(location.href);
+
+		if (matches) {
+			[, showTitle, seasonOrYear, subTitle] = matches;
+		}
+
+		const id = subTitle ?? '';
+		const title = showTitle?.split('-').join(' ') ?? '';
+		const episodeTitle = '';
+		const isSeason = subTitle?.lastIndexOf('s' + String(seasonOrYear) + 'a') ?? 0;
```
rafaelgomesxyz: It's not good to do conversions like this (`String(...)`). If you're trying to account for when `seasonOrYear` is null, isn't it better to check for that explicitly? You could just do `(seasonOrYear ?? '...')`.

Also, it's confusing that the variable is named `isSeason` but holds a number instead of a boolean. It's probably better to rename it to `seasonIndex` or something.

rafaelgomesxyz (changes requested):

src/streaming-services/vrtnu-be/VrtnuBeParser.ts
```diff
@@ -0,0 +1,77 @@
...
+		const episode =
+			isSeason > 0
+				? parseInt(subTitle?.substring(isSeason + ((seasonOrYear?.length ?? 0) + 2)) ?? '')
```
MrMamen: I think this is very confusing. Perhaps you could type an example of the format as a comment. It could be easier for future debugging if things stops working, or the format is changes as well. Another approach could be a regular expression, but I guess either works.

src/streaming-services/vrtnu-be/VrtnuBeApi.ts
```diff
@@ -0,0 +1,24 @@
+import { Api } from '../common/Api';
+import { registerApi } from '../common/common';
+
+// Define any types you need here.
```
MrMamen: These comment are really just for the template, and doesn't make so much sense here.

MrMamen (changes requested):

src/streaming-services/vrtnu-be/VrtnuBeParser.ts
rafaelgomesxyz: @ITSYV Looking back, you also used this logic for Disney Plus, which does make things a bit confusing. If you could use a regular expression for Disney Plus as well, like you did with the other services, I think it would be better. See [this](https://github.com/trakt-tools/universal-trakt-scrobbler/commit/81cef1ee778b802b76801a30189fcef5bb50b4e3#diff-d1fd825e3b0b628976a42f935a29148422a5aaa028b171432183015a39da4ddaR72-R75) vs [this](https://github.com/trakt-tools/universal-trakt-scrobbler/commit/436127ba79fc7189f22b10601e1287f1cc0f5a76#diff-382cf51bd6eb4ad35fbfdc29573c6df45c843378b66876bd59259461d22e3f0cR45-R54).

src/streaming-services/vrtnu-be/VrtnuBeApi.ts
rafaelgomesxyz: I've already merged the other PRs, which also had these comments, so I can just remove them myself later when I work on the general cleaning PR.

ITSYV: Oh sorry. I didn't really check those files anymore and indeed just copy/paste to the other branches

src/streaming-services/vrtnu-be/VrtnuBeParser.ts
ITSYV: The problem with this was that I got an error on " + seasonOrYear + "

`Operands of '+' operation must either be both strings or both numbers. Consider using a template literal  @typescript-eslint/restrict-plus-operands`

This seemed to be the quick fix for that. If you have a proposition to change it a better way, feel free :)

SeasonOrYear is never null by the way. There's always a value on the page / in the url.

seasonIndex is indeed a better name for this!

rafaelgomesxyz: But you define it as `let seasonOrYear: string | null = null;`, so even if, in practice, it will always eventually become a string, TypeScript doesn't know that. The solution should be what I suggested above: `(seasonOrYear ?? '')`

src/streaming-services/vrtnu-be/VrtnuBeParser.ts
ITSYV: "https://www.vrt.be/vrtnu/a-z/dertigers/3/dertigers-s3a1/#close"
VRT isn't using specific classes / id's for their elements and there is no specific logic in their buildup. So I can't scrape their page.

A regular expression could have also been working, but then I'm looping through possibly multiple matches, right?

Same for Disney+: I wasn't sure if there would be multiple "." or ":" after the season/episode info, as a title could have a ":" in it I guess. But a regex would work for sure :)

It's what you guys prefer, I can edit it if you want!

src/streaming-services/vrtnu-be/VrtnuBeParser.ts
ITSYV: Oh oops, I skipped the last part about "(seasonOrYear ?? '')"
I'll change it and also change the naming

src/streaming-services/vrtnu-be/VrtnuBeParser.ts
rafaelgomesxyz: > but then I'm looping through possibly multiple matches

I don't quite understand what you mean. Couldn't you easily modify the `/\/a-z\/(.+)\/(.+)\/(.+)\//` regex to directly get season/episode info like `/\/a-z\/(.+)\/(.+)\/(.+)-s(\d+)a(\d+)\//`, assuming it always ends in `-sXaY`?

ITSYV: Well, if it's a movie, it doesn't end with sXaY ...
That's the reason for the extra check

rafaelgomesxyz: I'm guessing movies don't end with that, so maybe `/\/a-z\/(.+)\/(.+)\/(.+?)(-s(\d+)a(\d+))?\//`?

rafaelgomesxyz: Yeah, commented before I saw your reply. I think the regex I posted above would work. Then you could do:

`[, showTitle, seasonOrYear, subTitle, seasonAndEpisode, season, episode]`

And just check for `seasonAndEpisode` to see if it's a show or movie.

ITSYV: That could be a possibility yes! 
But I'm also using this as the ID for the movie. 

If I'm using that logic, I would end up with this, right? : 

```
const matches = /\/a-z\/(.+)\/(.+)\/(.+?)(-s(\d+)a(\d+))?\//.exec(location.href);
if (matches) {
	[, showTitle, seasonOrYear, subTitle, season, episode] = matches;
}
const id = subTitle ?? '' + season ? '-s' + season : '' + episode ? 'a' + episode : '';
```

rafaelgomesxyz: `(-s(\d+)a(\d+))` is 3 capture groups (the order is always from the outer ones to the inner ones), so you'd end up with `[, showTitle, seasonOrYear, subTitle, seasonAndEpisode, season, episode]` where `seasonAndEpisode = '-s3a1'`, `season = '3'` and `episode = '1'`.

You could also put everything in another outer capture group: `((.+?)(-s(\d+)a(\d+))?)`. This would lead to `[, showTitle, seasonOrYear, id, subTitle, seasonAndEpisode, season, episode]` and you'd get the ID directly.

ITSYV: Learned something new again. Never used regex that advanced :)
I'll edit it to the last one 👍 

Do you like me to edit it for Disney+ too?

rafaelgomesxyz: If it's possible. I don't know the format for Disney+, maybe you can give us one example for that too?

ITSYV: Disney+ formats the subtitles / episode titles as: "S1: afl. 6 One World, One People"
I used "." because it's language dependent

Another question about the previous, you said "seasonAndEpisode, season, episode" but season and episode have to be of type "number". Should I name them: seasonStr and episodeStr and afterwards do a parseInt? (or use the same names and do a parseInt later on?)

ITSYV: Or can I use this when declaring: `let season: number | string | undefined = undefined;`
not sure if that's the way to go?

rafaelgomesxyz: > Should I name them: seasonStr and episodeStr and afterwards do a parseInt?

Yep, that's a lot better.

> Disney+ formats the subtitles / episode titles as: "S1: afl. 6 One World, One People"

Maybe something like `/(S(\d+).+?(\d+) )?(.+)/`? Then you'd have `[, seasonAndEpisode, seasonStr, episodeStr, subTitle]` and that would work for movies as well, assuming movies would be just 'One World, One People' without the beginning.

ITSYV: > Yep, that's a lot better.

alright, I'll edit it to episodeStr 👍 

> Maybe something like `/(S(\d+).+?(\d+) )?(.+)/`? Then you'd have `[, seasonAndEpisode, seasonStr, episodeStr, subTitle]` and that would work for movies as well, assuming movies would be just 'One World, One People' without the beginning.

Movies don't even have a subtitle, so this won't even be a problem

rafaelgomesxyz: Perfect. Only one more thing, if the 'S' changes based on language, it's probably best to just do: `/(.+?(\d+).+?(\d+) )?(.+)/` instead.

rafaelgomesxyz: And we just have to hope that there's no language where the episode comes before the season. haha

ITSYV: > Perfect. Only one more thing, if the 'S' changes based on language, it's probably best to just do: `/(.+?(\d+).+?(\d+) )?(.+)/` instead.

It could be that the "S" would change, not sure. In Dutch it's "seizoen", as "afl" is the abbreviation of "aflevering" = episode.

rafaelgomesxyz:
You can make the Disney+ changes here as well, no need for a new PR.

rafaelgomesxyz:
I think it would be beneficial to add an example like you did for this and Disney+ to the other 3 services as well, if you can.

ITSYV:
There you go! Disney+ code cleaned and also removed the obsolete comments from the other streaming services :)

rafaelgomesxyz:
> There you go! Disney+ code cleaned and also removed the obsolete comments from the other streaming services :)

Could you also add a format example to the other 3 services?

ITSYV:
> > There you go! Disney+ code cleaned and also removed the obsolete comments from the other streaming services :)
> 
> Could you also add a format example to the other 3 services?

I was already on it ;-) On it's way

rafaelgomesxyz (approved):
Awesome! Thanks again for all the new services!

ITSYV:
> Awesome! Thanks again for all the new services!

You're very welcome! I enjoyed traktflix for a long time and was happy to see that I could add Disney+ to this universal tracker myself! 
Though I'm not the best in Typescript, if I can help you guys with anything else.. or maybe test some stuff (like Disney+), let me know!

MrMamen (approved):
Great job! Too bad disney+ doesn't provide a viewable history. :-(

ITSYV:
> Great job! Too bad disney+ doesn't provide a viewable history. :-(

Uhu.. I've put in a feature request some months ago.. Got an automatic reply but never gotten back to me about it.
I hope it'll still be added in the future!

PR #74, opened 2021-05-28
Rewrite to use HTML-video-element when available.

### LINKED ISSUE #73: Progress always starts at 0
While this is ultimately up to each streaming service Event-file, there is currently no way to start a scrobble on anything other than 0. The scenario for this is when resuming a playback from the homepage of a streaming service. Calling `updateProgress `before `start`, will usually not work because `TraktItem `is not yet set. One possible solution would be to use `percentageWatched` on `Item`.

MrMamen:


MrMamen:
@ITSYV Check this out. It is no longer dependent of the controllers being visible. The title however is still needed as I found no other way.

ITSYV:
@MrMamen Oh, thanks!
I'll check it out for the other services, maybe it can be handled this way too!

rafaelgomesxyz:
Wow, I never thought of this. This is huge! It can also be done for Netflix, and I assume every other service as well. No more having to inject a script into the Netflix page to retrieve the session.

I think we should turn this PR into a PR for all services. I'll try to update Amazon Prime, HBO Go and Netflix.

src/streaming-services/disneyplus/DisneyplusParser.ts
```diff
@@ -20,44 +17,21 @@ class _DisneyplusParser implements ScrobbleParser {
...
-		} else if (playIcon && !pauseIcon) {
-			this.isPaused = true;
-		} else if (pauseIcon && !playIcon) {
-			this.isPaused = false;
-		}
-
-		const paused = this.isPaused;
-		const progress = this.parseProgress();
-		return { paused, progress };
+	getVideoElement = (): HTMLVideoElement => {
```
rafaelgomesxyz: Any particular reason why you put this in a separate function? Isn't it just a waste since it can be inlined?

MrMamen: Nah. Forgot to clean up. Started with just keeping the names identical. Didn't realise it was just used once. Will fix!

rafaelgomesxyz:
> I think we should turn this PR into a PR for all services. I'll try to update Amazon Prime, HBO Go and Netflix.

Actually, I think we can move this logic to the `ScrobbleEvents` and `ScrobbleParser` classes, to reuse it in every service, which would also make it easier to add new services. I'll work on something and push soon.

MrMamen:
Yeah. Was thinking the same thing thing. Have to make sure that all services uses standard Video element though.

MrMamen:
> > I think we should turn this PR into a PR for all services. I'll try to update Amazon Prime, HBO Go and Netflix.
> 
> Actually, I think we can move this logic to the `ScrobbleEvents` and `ScrobbleParser` classes, to reuse it in every service, which would also make it easier to add new services. I'll work on something and push soon.

Just to make sure, are you working on this branch @rafaelgomesxyz?

rafaelgomesxyz:
> Just to make sure, are you working on this branch @rafaelgomesxyz?

Yep

ITSYV:
> > Just to make sure, are you working on this branch @rafaelgomesxyz?
> 
> Yep

I'll wait for you to finish to check it out for the Belgian services

rafaelgomesxyz:
Sorry it's taking so long, I had to make a lot of changes. But I'm almost done, I just have to test Amazon Prime, HBO Go and Netflix a bit to make sure that they're working well. Will try to push the changes by tomorrow.

MrMamen:
> Sorry it's taking so long, I had to make a lot of changes. But I'm almost done, I just have to test Amazon Prime, HBO Go and Netflix a bit to make sure that they're working well. Will try to push the changes by tomorrow.

Yeah I figured, thanks for letting me know.

rafaelgomesxyz:
Alright, I think I'm done. Here's the list of changes:

* **Only use arrow class methods when needed:** When using arrow class methods, you can't call `super` from inherited classes, which is very useful, and it also creates the methods for each instance instead of having them in the prototype. The only reason why I had turned on the rule to enforce them was because using normal class methods as event listeners causes `this` to be unbound, but I didn't realize that there's already a rule for this: https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/unbound-method.md This rule works nicely for detecting those cases, so that we only have to use arrow class methods for event listeners.
* **Allow creating item with undefined ID and generate ID automatically:** Makes custom IDs more consistent. Still language-dependent though.
* **Allow creating item with undefined year and improve search result:** It's better than having to create with 0 when the year is missing.
* **Rename Item#percentageWatched to Item#progress for consistency**
* **Use undefined season and episode to represent collection instead of boolean:** Collections just mean "search by title instead of season and episode", so I think it's better to just do that when season and episode are undefined instead of having to rely on a boolean.
* **Improve messaging:** Now all the extension pages and the content script can handle messages and it's easier to setup new messages. Also, I figured out that by using port communication, it's possible to prevent duplicate script injections when using `browser.tabs.onUpdated` (#62), because it dispatches an event when the communication ends, which basically means that `webNavigation` is no longer needed, so it has been removed.
* **Improve watching popup:** It now shows a progress bar and indicates when it's paused.
* **Implement script injector:** Makes it easier to inject scripts when needed.
* **Improve scrobble detection and generalization:** The main changes here. Makes it easier to add new streaming services.

Amazon Prime, HBO Go and Netflix appear to be working well. All we have to do now is test the others, see if the video player method is working, and if that's the case, remove the current method being used.

src/streaming-services/common/ScrobbleEvents.ts
```diff
@@ -22,30 +51,77 @@ export abstract class ScrobbleEvents {
...
+		if (this.url !== newUrl) {
+			await this.onUrlChange(this.url, newUrl);
+			this.url = newUrl;
+		}
+
+		const playback = await this.parser.parsePlayback();
+		if (playback) {
+			const newProgress = playback.progress;
+			if (this.progress !== newProgress) {
+				await this.updateProgress(newProgress);
```
rafaelgomesxyz, line 65: Because it now updates the progress before starting the scrobble, I'm pretty sure this fixes #73.

rafaelgomesxyz: Wait, actually it still wouldn't work because `TraktItem` wouldn't be set. I'll fix it by getting the progress from `Item` before starting the scrobble.

rafaelgomesxyz: Alright, now it should work properly.

MrMamen (commented):
It seemed that I had to disable and re-enable disney+ for it to detect what I was playing. Even then when I exited the page, the icon is stuck in scrobble-mode, even if no scrobbing is going on.
For NRK I was trying to test the inject-script you had migrated to. But it doesn't seem to work at all. I'm planning on using the video-object here as well, but would make sure that the injection works as well. Not quite sure why, but as far as I can see it never even gets into the 'parsePlayback'-method.

ITSYV:
> It seemed that I had to disable and re-enable disney+ for it to detect what I was playing. 

I don't have that problem here. I just used the latest version from git (not master, but disney+ PR) and it was working fine.
VRTNu, VTMGo, GoPlay & Streamz are all working as well. Progress tested as well!

Nice touch with the progress bar in the view by the way!

rafaelgomesxyz:
> It seemed that I had to disable and re-enable disney+ for it to detect what I was playing. Even then when I exited the page, the icon is stuck in scrobble-mode, even if no scrobbing is going on.
> For NRK I was trying to test the inject-script you had migrated to. But it doesn't seem to work at all. I'm planning on using the video-object here as well, but would make sure that the injection works as well. Not quite sure why, but as far as I can see it never even gets into the 'parsePlayback'-method.

Hm that's weird.

It doesn't get into the main `parsePlayback` method or the `parsePlaybackFromInjectedScript` one? Do you get something if you uncomment the logging lines in the `ScrobbleParser` file?

rafaelgomesxyz:
> VRTNu, VTMGo, GoPlay & Streamz are all working as well. Progress tested as well!

But are they working because of the video player method or the old method? Because the old method is being used as fallback. If you uncomment the logging lines in the `ScrobbleParser` file you should be able to see which method it's using to parse it. Or if you remove the old method and it still works as well.

MrMamen:
> It doesn't get into the main `parsePlayback` method or the `parsePlaybackFromInjectedScript` one? Do you get something if you uncomment the logging lines in the `ScrobbleParser` file?

Nothing on NRK.  Nothing from any of these.

The Disney+-issue is probably not related to re-enabeling the scobbling, but by accident. There is an issue I'm looking into that can occur in certain situations it seems.

rafaelgomesxyz:
> The Disney+-issue is probably not related to re-enabeling the scobbling, but by accident. There is an issue I'm looking into that can occur in certain situations it seems.

I detected something similar, where the icon would turn to a scrobbling icon even though nothing was scrobbling because it failed to find the item. But I fixed that one with 0e8627c.

src/streaming-services/netflix/NetflixApi.ts
```diff
@@ -257,7 +248,7 @@ class _NetflixApi extends Api {
 			} while (!hasReachedEnd && itemsToLoad > 0);
 			if (historyItems.length > 0) {
 				const historyItemsWithMetadata = await this.getHistoryMetadata(historyItems);
-				items = historyItemsWithMetadata.map(this.parseHistoryItem);
+				items = historyItemsWithMetadata.map((historyItem) => this.parseHistoryItem(historyItem));
```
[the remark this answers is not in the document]
rafaelgomesxyz: It's because of the changes I made to only use arrow class methods when needed. I thought it was best to keep this method as a normal method (I kept only event listeners as arrow methods), so I had to change this so that `this` wouldn't get unbound.

src/streaming-services/sync-template/SyncTemplateApi.ts
```diff
@@ -13,11 +13,7 @@ class _SyncTemplateApi extends Api {
...
 
 	// This method should load the next page of history items and update the sync store for the service.
-	loadHistory = async (
-		itemsToLoad: number,
-		lastSync: number,
-		lastSyncId: string
-	): Promise<void> => {
+	async loadHistory(itemsToLoad: number, lastSync: number, lastSyncId: string): Promise<void> {
 		// The code could look like this.
 		const store = getSyncStore('sync-template');
```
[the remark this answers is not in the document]
rafaelgomesxyz: On #75 I moved this logic to the `Api` class to make it more generic. But yeah, I'll try to work on something so that we only have to enter the ID once.

src/streaming-services/common/ScrobbleEvents.ts
```diff
@@ -1,59 +1,131 @@
...
+				this.progress = newProgress;
+			}
+
+			try {
+				if (playback.isPaused && !this.isPaused) {
+					await this.pause();
+				} else if (!playback.isPaused && this.isPaused) {
```
MrMamen, line 72: It should not start a scrobble if item is not present. If the Item is retrieved later it will not (re)start the scrobble, and the only way to make it scrobble is to manually stop/start the player. Not sure if you want to prevent the playback from returning a value if item is not found, or if you want to include item in this file.

rafaelgomesxyz: Oh yeah, `parsePlayback` should return `null` if there's no item. I realize that I only checked for that before parsing the item, but not after to make sure that the item was actually parsed.

MrMamen:
I see that scrobbling on NRK stopped working on the last commit on master: [Implement automatic sync](https://github.com/trakt-tools/universal-trakt-scrobbler/commit/182321ca4ff043b2746dcedb8b7593b7cfee7c08). Since it is squashed the commit is quite large and difficult to find the specific change. Its just like the host-pattern doesn't match, but it is correct.

rafaelgomesxyz:
> I see that scrobbling on NRK stopped working on the last commit on master: [Implement automatic sync](https://github.com/trakt-tools/universal-trakt-scrobbler/commit/182321ca4ff043b2746dcedb8b7593b7cfee7c08). Since it is squashed the commit is quite large and difficult to find the specific change. Its just like the host-pattern doesn't match, but it is correct.

Well, the related PR has the commits unsquashed: #63

rafaelgomesxyz:
Ok I found the issue, it's here: https://github.com/trakt-tools/universal-trakt-scrobbler/blob/64f49004afbbbccefab377afe40ef333cac085af/src/modules/background/background.ts#L142

The way that the regex is being converted doesn't match the `tv.` in `tv.nrk.no`. I'll fix it.

rafaelgomesxyz:
> Ok I found the issue, it's here:
> 
> https://github.com/trakt-tools/universal-trakt-scrobbler/blob/64f49004afbbbccefab377afe40ef333cac085af/src/modules/background/background.ts#L142
> 
> The way that the regex is being converted doesn't match the `tv.` in `tv.nrk.no`. I'll fix it.

Should be fixed.

MrMamen:
Not sure if this is solvable without spamming trakt, but when seeking (changing the duration manually) it never gets updated to trakt, only locally. It might not be very important either, as long as the correct value is sent when stopping.

rafaelgomesxyz:
> I think this is ready for master.

I'm just waiting on confirmation that the Belgium services work with the video player method. I don't think @ITSYV has permission to push here, so I'd have to remove the old methods myself.

ITSYV:
> > I think this is ready for master.
> 
> I'm just waiting on confirmation that the Belgium services work with the video player method. I don't think @ITSYV has permission to push here, so I'd have to remove the old methods myself.

Do I need to make changes?
I just tested everything again with the updated code and all Belgian services are working!

rafaelgomesxyz:
> Do I need to make changes?
> I just tested everything again with the updated code and all Belgian services are working!

Just wanted to confirm that they're actually working through the video player method. The way that it's implemented, it falls back to the old method. So if you uncomment the logging lines in the `ScrobbleParser.ts` file you should be able to see which method it's using to parse it.

ITSYV:
> > Do I need to make changes?
> > I just tested everything again with the updated code and all Belgian services are working!
> 
> Just wanted to confirm that they're actually working through the video player method. The way that it's implemented, it falls back to the old method. So if you uncomment the logging lines in the `ScrobbleParser.ts` file you should be able to see which method it's using to parse it.

Streamz, GoPlay, VRTNu & VTMGo:
Playback parsed from video player!

rafaelgomesxyz (approved):
Done!

PR #75, opened 2021-06-06
Rewrite to improve generalization

rafaelgomesxyz:
- [x] **Improve sync generalization:** move common logic to the `Api` class to make adding new services easier
- [x] **Improve general generalization:** to prevent having to repeat the service ID
- [x] **Isolate services:** make sure that service implementations are isolated to their own folder, to prevent merge conflicts when merging PRs for different services
- [x] **Implement CLI script to create services:** to make it much easier to add new services

Will rebace once #74 is merged.


rafaelgomesxyz:
@MrMamen I generalized things even more here.

Now each service implementation is isolated to its own folder, while Webpack takes care of generating the content script files and filling the necessary information in the `streaming-services.ts` and `apis.ts` files.

I also made a CLI script to create services. In order to test it:

```
cd create-streaming-service
npm link
cd ..
npx create-streaming-service
```

What do you think? If it all looks ok to you, I'll move the `create-streaming-service` folder to its own repo and install it as a dependency from NPM.

src/streaming-services/netflix/NetflixApi.ts
```diff
@@ -56,7 +55,7 @@ export interface NetflixHistoryResponse {
 
 export type NetflixHistoryItem = NetflixHistoryShowItem | NetflixHistoryMovieItem;
 
-export interface NetflixHistoryShowItem {
+export interface NetflixHistoryShowItem extends HistoryItem {
```
MrMamen: Are you sure that this won't open up the interface to basically anything?

MrMamen (commented):
Wow. Cool. I'll test this is a little while. Must be a lot of work to do these changes as the number of streaming services are increasing.

src/common/Events.ts
```diff
@@ -89,7 +88,7 @@ export interface OptionsChangeData<K extends keyof StorageValuesOptions> {
 	value: StorageValuesOptions[K];
 }
 
-export type StreamingServiceOptionsChangeData<K extends StreamingServiceId> = {
+export type StreamingServiceOptionsChangeData<K extends string> = {
```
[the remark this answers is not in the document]
rafaelgomesxyz: Yeah, I just replaced all occurrences of `StreamingServiceId` with `string`. Will check more closely to remove those redundancies.

src/streaming-services/netflix/NetflixApi.ts
rafaelgomesxyz: Not sure. I guess we could use `unknown` instead of `HistoryItem` in the `Api` class, because the type doesn't really matter there.

rafaelgomesxyz:
> Must be a lot of work to do these changes as the number of streaming services are increasing.

Hopefully it will save us work in the future.

That's actually another thing that I wanted to talk about. As the number of streaming services increases, so does the extension size, and considering that each service has its own content script file, this could lead to a rather large size later on. Do you think that we should just have one content script file for all services? This would increase the size of the file that has to be injected, but it would reduce the size overall.

MrMamen:
> That's actually another thing that I wanted to talk about. As the number of streaming services increases, so does the extension size, and considering that each service has its own content script file, this could lead to a rather large size later on. Do you think that we should just have one content script file for all services? This would increase the size of the file that has to be injected, but it would reduce the size overall.

Good question. I'm a bit worried about including a lot of unused code in each script. At least if they start interfering with each other. But I do see that the size seems alarmingly big already.  Not sure if it matters much that the size is large though. Might eat memory I suppose, and the initial download might be slower.

Good work with the script! Tested it now.

rafaelgomesxyz:
Moved the CLI to a separate repo (https://github.com/trakt-tools/cli) and added a command to update a streaming service as well.

So this PR is ready.

MrMamen (approved):

PR #77, opened 2021-06-15
Update dependencies

rafaelgomesxyz:
Webpack 5 changed how JSONs are imported, so I just decided to replace the service JSONs with TypeScript files. I think this is better anyway, for correctly detecting the types. I'll have to update the CLI as well.

I had to delete my whole `node_modules` folder and reinstall everything to make the build work.


MrMamen:
Even after deleting node_modules I get A LOT of errors like this:

> ERROR in ./node_modules/@material-ui/core/node_modules/@babel/runtime/helpers/esm/inherits.js 1:0-46
Module not found: Error: Can't resolve './setPrototypeOf' in '\universal-trakt-sync\node_modules\@material-ui\core\node_modules\@babel\runtime\helpers\esm'
Did you mean 'setPrototypeOf.js'?
BREAKING CHANGE: The request './setPrototypeOf' failed to resolve only because it was resolved as fully specified
(probably because the origin is a '*.mjs' file or a '*.js' file where the package.json contains '"type": "module"').
The extension in the request is mandatory for it to be fully specified.
Add the extension to the request.
 @ ./node_modules/@material-ui/core/esm/RootRef/RootRef.js 3:0-60 50:2-11
 @ ./node_modules/@material-ui/core/esm/RootRef/index.js 1:0-36 1:0-36
 @ ./node_modules/@material-ui/core/esm/index.js 175:0-47 175:0-47 176:0-26 176:0-26
 @ ./src/modules/history/HistoryApp.tsx 2:0-46 55:39-48
 @ ./src/modules/history/history.tsx 9:0-42 16:105-115

rafaelgomesxyz:
Those were the errors I was getting as well. Oh also make sure to delete
the `package-lock.json` or `pnpm-lock.yaml` file.

MrMamen (approved):
package-lock.json should be excluded from .gitignore BTW. The purpose of the file is to ensure everybody gets the same versions of every dependency.
Besides that, this solves a lot of warnings, I got!
I like the idea to use ts-files instead of json as well.

rafaelgomesxyz:
Are you ok with using pnpm instead of npm? The reason why I added `package-lock.json` to `.gitignore` is because I personally use pnpm (so does the CI), since it's faster and uses less disk space.

MrMamen:
oh. right. I personally use Yarn. Which was why I initially got this error I think. I completely forgot about the yarn.lock-file. If you recommend pnpm, i'll be happy to test it out.

rafaelgomesxyz:
I'm down to switch to Yarn too, we just have to choose one and ensure that it's used everywhere.

I've never used Yarn, but the reason why I prefer pnpm over npm is because if you have, for example, 50 repos that all use the same package and same version, with npm it will create 50 copies of that package in each repo's `node_modules`, while with pnpm it will store all packages in a single central folder and then use symbolic links in each repo's `node_modules`, so you only end up with 1 copy of the package that is referenced across all repos.

MrMamen:
It really just important whenever anyone updates the package.json, and since that is usually you I leave that decision to you. For anyone not updating package.json, npm, pnpm, and yarn would probably work anyways.

rafaelgomesxyz:
Ok, I'll keep pnpm then, since it's already using it, so no changes are needed.

PR #78, opened 2021-06-16
Support scrobbling in Viaplay

MrMamen:


src/streaming-services/viaplay/ViaplayApi.ts
```diff
@@ -89,19 +97,29 @@ class _ViaplayApi extends Api {
 	}
 
 	async activate() {
-		const response = await fetch(this.INITIAL_URL);
-		const host = response.url.split('//')[1];
+		let host;
+		let isLoggedIn = false;
+		if (location.hostname.includes('viaplay')) {
```
rafaelgomesxyz, line 102: This is a bit confusing. How does the URL having `viaplay` determine whether the user is logged in or not? Specially since, when scrobbling, this should always be true because the user would always be in a Viaplay page.

MrMamen: Yes. It is only needed from history sync, not from scribbling. Still need to run activate to get the urls though. I'm open for suggestions for alternative implementations.

rafaelgomesxyz: I was just referring to the variable being called `isLoggedIn`. Because that doesn't really mean that the user is logged in, does it? In the sense of them having entered their username/password?

I say this because I'm working on a PR that will add an `isLoggedIn` method to the `Api` class and will show a message to the user if they're not logged in to the service that they want to sync.

Wouldn't something like `isInViaplayPage` (or a better name) describe it better?

MrMamen: You are right. IsAuthorized perhaps?

MrMamen: This could also be a input parameter to the method. But isScrobbling could also work I guess.

rafaelgomesxyz: I like `isAuthorized`, but any of those is fine.

rafaelgomesxyz (approved):

Issue #44, opened 2020-10-27
Fix Netflix sync time


rafaelgomesxyz:
The Netflix watched date is always when the user began watching, not when they finished. We should change that by adding the length of the episode/movie to the watch date, so it's more accurate.

MrMamen:
Possible as an user option?

rafaelgomesxyz:
Well, I can't see why anyone would want it to behave like it does currently, since the watch time is always wrong. But sure, we can add it as an option, enabled by default.

MrMamen:
I think this is a case with pretty much all streaming services. It records the last time the play-button was clicked, or similar. The problem is of course if you watch a part of the title at one point, and resumes at another time. Not sure if it's worth making a lot of UI for, but at least it is a concern.
Also there are a few similar off situations.
If you use trakt "add on release date"-feature, it would record the start time. Scrobling and checkins would record the end time. I guess it is a matter of definition, and not sure if it is important. I don't have any strong opinions of this.

Issue #72, opened 2021-05-25
Sync History


ITSYV:
It would be a good idea to add a message to the sync history page (where you select the streaming services) that you have to be logged in to the service before you can sync.

And maybe a good idea to add a button to the streaming service sync page to go back when an error message is shown.
Also loadHistoryError should maybe be accompanied with why this error could have happened (maybe also a msg that you possibly aren't logged in and a link to the streaming service?)

rafaelgomesxyz:
Yep, I have plans to check for login when loading the history.

Issue #46, opened 2020-10-27
Make changes for rate limiting


rafaelgomesxyz:
Trakt announced they'll start rate limiting requests to the API today (https://trakt.docs.apiary.io/#introduction/rate-limiting). The new limits are:

* `POST` to `/sync/history/*` - 2 calls every second
* All other API methods - 1000 calls every 5 minutes

We should make sure our requests don't exceed this.

We already have a local cache, but it might be a good idea to setup a Google sheet on top of that, like the one we already have for TMDB images.

PR #80, opened 2021-06-26
Various bug fixes and enhancements

rafaelgomesxyz:
This will be another big PR (hopefully the last one before the release). I hope to go over everything left on my to-do list.


rafaelgomesxyz:
@MrMamen Google offers a free tier with their cloud products, so I think it's better to store the database there instead of the Google Sheet. I separated a subdomain in my domain and was making some tests, and it seems to work quite well. Example: https://uts.rafaelgomes.xyz/api/correction-suggestions?ids=netflix_80103820,netflix_80103821 If it takes a while to load, it's because the instance was suspended (Google suspends instances that are idle for too long to save resources).

The free tier offers 28 instance hours per day (since we'll only use one instance, we'll only need 24 hours at most), and 50.000 database writes / 20.000 database reads per day (which I think should be more than enough).

If that's ok with you, I'll add `uts.rafaelgomes.xyz` as a required permission and make the appropriate changes.

MrMamen:
Sure go ahead. It might be slightly harder to check the list, but it will probably be a smoother experience for the users.

rafaelgomesxyz:
For us, not really. I added you to the project, so if you go to https://console.cloud.google.com/firestore/data/?project=trakt-tools-uts you should be able to see the database.

rafaelgomesxyz:
Will start pushing commits tomorrow, since there are a lot of changes, in case you want to start reviewing them.

I think we can finally release it this month. This PR will fix #44, #46 and #72, and the only other issue that would be nice to fix before release is #47, but I think I can do that one quickly after I'm done with this PR. Will see how much I can do this weekend.

rafaelgomesxyz:
Had to deal with something last week, so didn't have time to work on it, but now I do.

I decided to make separate PRs in the end, to separate things better. First one is up: #81

rafaelgomesxyz:
@MrMamen I'm currently working on a PR to replace history pages with infinite scrolling. Just wondering if you're ok with that or if you think we should keep the pages.

I think the infinite scrolling is better because we don't have to worry about each page, and it also makes it easier to load more items if some items have been filtered. Right now, even if `itemsToLoad` is set to 10 and, let's say, 5 items were filtered because they're already synced, the page will display only 5 items. With infinite scrolling, we can load items on demand, so it doesn't matter how many items were filtered, it will always load more.

In that PR, I'm also removing the `itemsToLoad` option and making it always load 10, because since it's infinite scrolling, it doesn't make sense to allow the user to specify how much to load each time, and I think 10 is a good number. Let me know if you think we should keep this option as well.

MrMamen:
@rafaelgomesxyz I think infinite scrolling is great.

rafaelgomesxyz:
> I think we can finally release it this month.

And two months passed... lol

I'll try to finish the last PRs this next week. Only things I have left to do:

- [x] final dependency update, to get rid of some deprecated or outdated packages (Material-UI v4 to v5, moment to date-fns, etc.)
- [ ] better typings (generic request response, split `Item` into `Episode` and `Movie`, remove all `never` types, etc.)
- [ ] better errors (make sure that every error is properly dealt with)
- [ ] fix any known remaining bugs (this PR)
- [ ] prepare for traktflix migration
- [ ] release

MrMamen:
Any progress on this one?

rafaelgomesxyz:
Still working on it, but been a bit busy recently. Almost done with the better typings PR.

PR #82, opened 2021-07-17
Move database from Google Sheet to Google Cloud

rafaelgomesxyz:
The renames are either because they're shorter and get the point across or because of consistency.

The virtualized list changes the items while the user scrolls, to improve performance with large lists (this will later be used in the history page as well).

Everything else should be straightforward.

### DIFF
--- a/src/_locales/en/messages.json
+++ b/src/_locales/en/messages.json
@@ -114,11 +114,11 @@
 		"description": "This is the verb, as in 'to correct'.",
 		"message": "Correct"
 	},
-	"correctWrongItem": {
-		"message": "Correct wrong item"
+	"correctItem": {
+		"message": "Correct item"
 	},
-	"correctWrongItemFailed": {
-		"message": "Failed to correct wrong item."
+	"correctItemFailed": {
+		"message": "Failed to correct item."
 	},
 	"couldNotScrobble": {
 		"message": "Could not scrobble."
@@ -403,7 +403,7 @@
 	"watched": {
 		"message": "Watched On"
 	},
-	"wrongItemDialogContent": {
+	"correctionDialogContent": {
 		"message": "Enter the correct Trakt URL for $TITLE$ in the field below.",
 		"placeholders": {
 			"title": {
@@ -411,7 +411,7 @@
 			}
 		}
 	},
-	"wrongItemDialogContentWithSuggestions": {
+	"correctionDialogContentWithSuggestions": {
 		"message": "Enter the correct Trakt URL for $TITLE$ in the field below or use one of the suggestions.",
 		"placeholders": {
 			"title": {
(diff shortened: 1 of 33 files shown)


rafaelgomesxyz:
There are also a couple new messages to translate.

src/_locales/nb/messages.json
```diff
@@ -114,11 +114,11 @@
 		"description": "This is the verb, as in 'to correct'.",
 		"message": "Korriger"
 	},
-	"correctWrongItem": {
-		"message": "Korriger feilaktig element"
+	"correctItem": {
+		"message": "Korriger element"
```
MrMamen, line 118: Nice. Got lucky there!

src/models/TraktItem.ts
```diff
@@ -20,7 +20,7 @@ export interface TraktItemBase {
 
 export interface SavedTraktItem extends TraktItemBase {
 	releaseDate?: number;
-	watchedAt?: number;
+	watchedAt?: number | null;
```
MrMamen, line 23: Do we need to keep both undefined and null?

src/_locales/en/messages.json
```diff
@@ -405,21 +403,26 @@
 	"watched": {
 		"message": "Watched On"
 	},
-	"wrongItemDialogContent": {
+	"correctionDialogContent": {
```
MrMamen, line 406: Perhaps you should sort this with new order now. (possible a job that can be automated)

MrMamen (commented):
Not tested as I am not at my computer for a while, but have reviewed the code. Looks good.

src/models/TraktItem.ts
rafaelgomesxyz: Yes, because there's a distinction between the two:

* `undefined` means that we have not tried to retrieve the data yet
* `null` means that we have tried to retrieve the data and it was unavailable or nonexistent

Based on this, we can either show "Loading history..." if it's `undefined` or "Not Watched" if it's `null`.

I think this works best, as I don't see how else we would make this distinction. This distinction is also used in many other parts of the extension that involve retrieving data from somewhere, like when having to retrieve suggestions or images.

src/_locales/en/messages.json
rafaelgomesxyz: I have this for another PR as well.

src/models/TraktItem.ts
MrMamen: Ahh! Thanks for clarifying!

MrMamen (approved):

src/_locales/nb/messages.json
```diff
@@ -203,6 +203,9 @@
 	"loadTraktHistoryError": {
 		"message": "Klarte ikke å laste inn Trakt historikken."
 	},
+	"loadingHistory": {
+		"message": "Loading history"
```
MrMamen: "Laster inn historikk"

src/_locales/nb/messages.json
```diff
@@ -405,21 +403,26 @@
...
 		"placeholders": {
 			"title": {
 				"content": "$1"
 			}
 		}
 	},
-	"wrongItemDialogContentSuggestions": {
-		"message": "Korrigeringsforslag (trykk for å bruke):"
+	"correctionDialogContentWithSuggestions": {
+		"message": "Enter the correct Trakt URL for $TITLE$ in the field below or use one of the suggestions.",
```
MrMamen: Skriv inn riktig Trakt URL til $TITLE$ i feltet nedenfor, eller bruk en av forslagene.

MrMamen (dismissed):
Translated in comments

MrMamen (approved):

rafaelgomesxyz:
> Translated in comments

Thanks, done!

PR #85, opened 2021-07-30
History improvements

### LINKED ISSUE #44: Fix Netflix sync time

### LINKED ISSUE #72: Sync History

rafaelgomesxyz:
* **Use esModuleInterop:** fixes some issues when importing default modules
* **Move background page logic to the appropriate places:** so that it's reusable, also makes it so that you don't have to refresh the page when changing options for things to take effect
* **Allow some events to be dispatched globally:** so we don't have to add specific messages for them, also makes communication between extension and content scripts easier
* **Improve options:** makes some components reusable, also prevents unecessary re-renders
* **Improve sync options:** same as above, but for sync options
* **Implement infinite scrolling:** better than having to deal with pages, also makes things flow better
* **Use contexts instead of passing props in some cases:** better than having to pass props down through several components if it makes more sense to have them in a context
* **Make it easier to use react-devtools:** just a dev change, for profiling, this is what I used to improve the options
* **Update enabled services live in the history page:** so that you don't have to refresh the page after enabling/disabling a service to see it in the list
* **Fix Netflix sync time:** I decided not to add the duration here, because, as mentioned in #44, it can get messy if the user stops watching something and returns later, so I think it's best to just leave it the way it is (fixes #44)
* **Check if user is logged in to the service before opening sync page:** improves user experience, also shows the name of the current profile in the sidebar, which can let the user know if they're in the right profile (fixes #72)
* **Return to home page on error:** improves user experience


rafaelgomesxyz:
@MrMamen Some new messages here as well.

It would be nice if you could test the new login check for NRK and Viaplay. I couldn't fully test it with a VPN.

Also, the infinite scrolling might need some more testing, but as far as I can tell it's working fine. You'll notice that options are disabled while items are loading, and you can't select items while they're loading either. This is because the user can now interact with the page while the next one is loading, so this behavior prevents conflicts.

I think I have a couple more PRs after this one before we can possibly release it.

src/common/Errors.ts
```diff
@@ -33,6 +24,34 @@ class _Errors {
...
+			this.rollbar = new Rollbar({
+				accessToken: secrets.rollbarToken,
+				autoInstrument: {
+					network: false, // Do not set to true on Firefox (see https://github.com/rollbar/rollbar.js/issues/638).
+				},
+				captureIp: false,
+				captureUncaught: true,
+				captureUnhandledRejections: true,
+				payload: {
+					environment: 'production',
```
MrMamen: I think it would be better to read this from an enviroment variable. So that the same rollbarToken could be used for both production and development. (Of course not important improvement.)

src/common/BrowserStorage.ts
```diff
@@ -158,36 +159,73 @@ export type CorrectItem = {
...
+	| SwitchOptionDetails<T, K>
+	| TextFieldOptionDetails<T, K>
+	| NumericTextFieldOptionDetails<T, K>
+	| CustomOptionDetails<T, K>;
+
+export type OptionDetailsByType<
+	T,
+	U extends OptionDetails<T, K>['type'],
+	K extends keyof T = keyof T
+> = U extends 'select'
```
MrMamen, line 188: Yikes! I guess you used some time in getting this right?

src/_locales/nb/messages.json
```diff
@@ -265,12 +262,12 @@
...
 		"message": "Legg inn en gyldig dato!"
 	},
-	"nextPage": {
-		"message": "Neste side"
-	},
 	"no": {
 		"message": "Nei"
 	},
+	"noItemsSelected": {
+		"message": "No items are selected. Select some items and try again."
```
MrMamen: Ingen elementer er valgt. Kryss av for noen elementer og prøv på nytt.

src/_locales/nb/messages.json
```diff
@@ -395,6 +389,9 @@
 	"sync": {
 		"message": "Synkroniser historikk"
 	},
+	"syncing": {
+		"message": "Syncing"
```
MrMamen: Synkroniserer

MrMamen (commented):
I just had a brief look, but I'll have more time to review and test this in a week's time.

src/common/BrowserStorage.ts
rafaelgomesxyz: A little. Do you know of a better way to deal with this?

MrMamen: Nope. It's good, just glad I didn't have to figure it out!

src/_locales/nb/messages.json
```diff
@@ -280,6 +277,12 @@
 	"notFound": {
 		"message": "Ikke funnet"
 	},
+	"notLoggedIn": {
+		"message": "Not Logged In"
```
MrMamen: Ikke logget inn

src/_locales/nb/messages.json
```diff
@@ -280,6 +277,12 @@
...
+	},
+	"notLoggedInDescription": {
+		"message": "Looks like you're not logged in to this service. Please go to the page below and log in before trying to sync."
```
MrMamen: Det ser ut som du ikke er innlogget på denne tjenesten. Forsøk å gå inn på siden nedenfor og gjør en innlogging før du forsøker å synkronisere.

src/_locales/nb/messages.json
```diff
@@ -295,8 +298,11 @@
 	"options": {
 		"message": "Alternativer"
 	},
-	"previousPage": {
-		"message": "Forrige side"
+	"profile": {
+		"message": "Profile"
```
MrMamen: Profil

src/_locales/nb/messages.json
```diff
@@ -295,8 +298,11 @@
...
+	},
+	"profileDescription": {
+		"message": "This is the current profile. If this is not the profile that you want to sync, switch profiles in the service page."
```
MrMamen: Dette er den valgte profilen. Dersom dette ikke er profilen du ønsker å synkronisere, forsøk å bytte profilen inne på siden til strømmetjenesten.

src/_locales/nb/messages.json
```diff
@@ -395,24 +401,27 @@
...
 		"message": "Velg ønsket tema."
 	},
-	"themeName": {
-		"message": "Tema"
-	},
 	"toggleAll": {
 		"message": "Veksle"
 	},
+	"unknown": {
+		"message": "Unknown"
```
MrMamen: Ukjent

MrMamen (commented):
I'm currently testing and reviewing this.  The translations are added for now.

src/modules/history/components/HistoryOptionsList.tsx
```diff
@@ -53,6 +55,16 @@ export const HistoryOptionsList: React.FC = () => {
 	return (
 		<Drawer classes={{ paper: 'history-options-sidebar' }} anchor="left" variant="permanent">
 			<Toolbar />
+			<Tooltip title={I18N.translate('profileDescription')}>
+				<Box textAlign="center">
+					<Typography variant="overline">
+						<Box display="inline" fontWeight="fontWeightBold">
+							{I18N.translate('profile')}:
+						</Box>{' '}
+						{api?.session?.profileName || I18N.translate('unknown')}
```
MrMamen: I think this should be highlighted a lot! Perhaps even moved to the main window.

MrMamen:
I've noticed that if the element is not found (which is quite common on NRK), The trakt-box never gives the user the option to correct the item.
image

Not really a problem with this PR, but I've noticed that history sync doesn't work without also enabling scrobbling.  Seems to be a problem with of CORS on auth for getting the token.

src/modules/history/components/HistoryOptionsList.tsx
rafaelgomesxyz: Any suggestions? Moved to above the items?

rafaelgomesxyz:
> Not really a problem with this PR, but I've noticed that history sync doesn't work without also enabling scrobbling. Seems to be a problem with of CORS on auth for getting the token.

Yeah, the extension should be asking for host permissions whether the user is enabling scrobble or sync.

src/modules/history/components/HistoryOptionsList.tsx
MrMamen: Yeah, above the items. Preferable in the blue header bar. Centered: So it is always on top. The text could be something like: "History sync for {streamingService}, user: {profileName}

rafaelgomesxyz: Done. I used "profile" instead of "user" because it's already translated.

rafaelgomesxyz:
The new commit adds a new message ("for").

rafaelgomesxyz:
> > Not really a problem with this PR, but I've noticed that history sync doesn't work without also enabling scrobbling. Seems to be a problem with of CORS on auth for getting the token.
> 
> Yeah, the extension should be asking for host permissions whether the user is enabling scrobble or sync.

Hmm, actually, it already does. I think this is only a problem if you disable scrobble, because it removes the permission.

@MrMamen Can you confirm? After you disable scrobble, you can disable and re-enable the sync to re-add the permission.

rafaelgomesxyz:
Added just one more commit to implement a history cache, as this was the only cache that was missing.

It works like this:

1. The first history page is always fetched
2. The first item in the page is compared to the first item in the cache
3. If the items are different, the cache is cleared, otherwise the cache is used
4. When the user reaches the last item in the cache, history pages start being fetched again, using the next page/URL saved in the cache
5. Every time a new history page is fetched, the next page/URL and the items are saved in the cache

If no new items appear in the first history page, the cache is cleared after 24 hours.

This should prevent unnecessary requests to history pages, and also allow users to e.g. load pages 1-4 and then quickly start loading from page 5 later.

I only tested it for Netflix, so it would be nice to test it for NRK and Viaplay (especially since they use next URL instead of next page), but it appears to be working well.

src/modules/history/components/HistoryHeader.tsx
```diff
@@ -64,6 +66,30 @@ export const HistoryHeader: React.FC = () => {
 							</Button>
 						</>
 					}
+					middle={
+						<Typography variant="overline">
+							{serviceId !== null ? (
+								service ? (
+									<>
+										{I18N.translate('history')} {I18N.translate('for')}{' '}
```
MrMamen: I can't see any reason for not combining these two i18n-strings into one. Besides prepositions are often hard to translate because they are dependent on the context. (In this case the Norwegian translation should be "for", identical as the English).

MrMamen (dismissed):
Seems to be working now, but I would advice translations to stay as whole sentences rather than words when possible.

src/modules/history/components/HistoryHeader.tsx
rafaelgomesxyz: I thought it would be repetitive to have "history" and "historyFor", but you're right, it's best to use full sentences. Just pushed a commit that merges them.

MrMamen (approved):
Well done!

PR #90, opened 2021-08-16
Manage translations through Crowdin

rafaelgomesxyz:
I think it's better to manage translations separately instead of blocking PRs until they're translated, because that could potentially slow down reviews a lot while we wait for translations. This also removes the need for my `eslint-plugin-i18n-json` fork, because identical keys are handled at Crowdin.

https://crowdin.com/project/universal-trakt-scrobbler


MrMamen (dismissed):
Very good idea. This has been a concern for me as well. No experience with crowdin but this could perhaps make it easier for more translations as well. Perhaps invite people to translate in the readme?

rafaelgomesxyz:
Done.

Also, it's reporting some possible spellcheck errors. Could you check those? I sent you a manager request through email, but if you didn't receive it I can try sending it again. You can check them by clicking the three dots next to a file and selecting "Proofread".

MrMamen:
I did not get the manager request. But I did create a account and added suggestions to errors I found.

rafaelgomesxyz:
Sent the request again through Crowdin this time.

Only issue with this is that it doesn't look like it gives credit to the translator (all PRs are authored by me). Not sure if there's a setting to change that, but for now I'm adding "Co-Authored" manually.

rafaelgomesxyz:
Ok, I created a bot account because I think it's better to have that as the author of the PRs instead of me. So I had to remake the project on Crowdin with the bot as the owner and sent you a new request.

PR #97, opened 2021-08-25
Replace HBO Go with HBO Max

### LINKED ISSUE #86: HBO Max Support
I was looking for a way to scrobble HBO Max and this is the closest thing I could find. Would love to see HBO Max support in this project, keep up the good work!

rafaelgomesxyz:


MrMamen (approved):
Cool. HBO Max is supposed to be launched in Europe soon as well. Replacing a local version, here as well.

PR #98, opened 2021-08-25
Add sync for Amazon Prime

rafaelgomesxyz:


MrMamen (approved):
WOW. Did prime suddenly get history, or did we simply not find it last we checked?
You changed some HBO-files here as well. Was it just a forgotten cleanup?

rafaelgomesxyz:
> WOW. Did prime suddenly get history, or did we simply not find it last we checked?

It used to be a very basic history. For TV shows it would only say the show title instead of the specific season and episode. I'm not sure when they improved it, but I recently noticed that it now lists all the episodes that were watched, so I decided to investigate the requests and found the endpoint.

> You changed some HBO-files here as well. Was it just a forgotten cleanup?

Yeah, just related to no longer having to convert the ID, since Prime IDs also have special characters.

PR #100, opened 2021-09-16
Temporary fix Amazon Prime EU (+global?) - removal of api usage

ITSYV:
Added a service for Amazon Prime EU. Just using the same way the other services work.
Still needs fix for history sync, but not sure how we can solve this so both the EU and American version works

By the way, is there an option to add names to the "suggestion" list?
Shameless is called "Shameless: the complete series", but can't be found that way. I fixed it at this moment with splitting the title. But that's not best practice :)

### DIFF
--- a/src/services/amazon-prime-eu/AmazonPrimeEuApi.ts
+++ b/src/services/amazon-prime-eu/AmazonPrimeEuApi.ts
@@ -1,0 +1,10 @@
+import { AmazonPrimeEuService } from '@/amazon-prime-eu/AmazonPrimeEuService';
+import { ServiceApi } from '@apis/ServiceApi';
+
+class _AmazonPrimeEuApi extends ServiceApi {
+	constructor() {
+		super(AmazonPrimeEuService.id);
+	}
+}
+
+export const AmazonPrimeEuApi = new _AmazonPrimeEuApi();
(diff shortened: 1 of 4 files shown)


rafaelgomesxyz:
I don't like the idea of adding separate versions of a service for different regions. It should be just one for all. There's a way to make it work by taking advantage of the fact that it uses the next method as fallback when a method doesn't work. I can work on it when I have some time.

> By the way, is there an option to add names to the "suggestion" list?

Not entirely sure what you mean. You can't correct it by providing the URL to the correct item?

ITSYV:
You're totally right. I just made a separate one for the ease of use at this moment. I wasn't sure what your idea is with the Prime API.
Just wanted to share :)

> Not entirely sure what you mean. You can't correct it by providing the URL to the correct item?
No, it just doesn't recognize the item if i remove the split function. So I can't provide a correct url :)

MrMamen:
What is the status of this branch? Is it relevant?

ITSYV:
Not really.
The code could be used to make a fallback for the non-api version, and the API version should be extended with regional API url's. But I lack the experience with typescript to do this myself

rafaelgomesxyz:
Yeah, I'll work on this soon.

PR #101, opened 2021-09-18
Update dependencies

rafaelgomesxyz:
Some notes:

- Update Material-UI v4 to MUI v5:
  - style definitions were changed from Sass to the new `sx` inline method
  - as a result, new components were created to reuse styles
  - an ESLint plugin was created to order the CSS properties for consistency (https://www.npmjs.com/package/eslint-plugin-mui)
  - now using [Pug](https://github.com/pugjs/pug) to generate HTML files
- Replace legacy moment with date-fns:
  - since date-fns uses the native `Date`, this makes it easier to deal with dates
- Remove prop-types:
  - because it leads to too much code duplication, since we already type the props with TypeScript
- Use new JSX transform:
  - https://reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html


MrMamen (dismissed):

MrMamen:
Well done!

isleshocky77:
@rafaelgomesxyz This PR seemed to have broken the trakt login process for me. When you first try to use the plugin, `auth = null` and thus it never fires off the `LOGIN_FAILED` message to show the `Login with Trakt.tv` button on LoginPage - just a loading spinner which never ends.

PR #106, opened 2021-10-30
Fix invalid dateformat in sync

### LINKED ISSUE #105: Sync history added with current date/time instead of watched date/time
This is probably because Trakt now uses UTC datetime instead of unix time stamp: https://trakt.docs.apiary.io/#reference/sync/add-to-history/add-items-to-watched-history

> Send a `watched_at` UTC datetime to mark items as watched in the past. This is useful for syncing past watches from a media center.

I did a quick test by sending the same request, but changed `watched_at` to UTC datetime format instead and it worked.

MrMamen:
Closes #105

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,7 +44,6 @@
 		"@mui/system": "^5.0.0",
 		"axios": "^0.21.4",
 		"date-fns": "^2.24.0",
-		"date-fns-tz": "^1.1.6",
 		"deepmerge": "^4.2.2",
 		"history": "^4.10.1",
 		"react": "^17.0.2",
--- a/package.json
+++ b/package.json
@@ -44,6 +44,7 @@
 		"@mui/system": "^5.0.0",
 		"axios": "^0.21.4",
 		"date-fns": "^2.24.0",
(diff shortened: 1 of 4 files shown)


rafaelgomesxyz:
Is `date-fns-tz` actually needed? I thought this bug was just a format issue, because as far as I know the dates are already in UTC timezone.

MrMamen:
I was not able to get the correct format in UTC without it. But I might have missed something in the API.

rafaelgomesxyz:
Ok, will take a look soon.

MrMamen:
When using `dateFns.formatISO` I always get the date in the local timezone. While the time is correct, the trakt API doesn't accept this. `date-fns-tz` was the only way I was able to get past this.

rafaelgomesxyz:
But why use `dateFns.formatISO`? Couldn't you just get the UTC format by using the native `Date.toISOString`? So you could do `new Date(unix * 1e3).toISOString()` and get the desired format, no?

rafaelgomesxyz (approved):

PR #127, opened 2022-03-07
Fix for HBO Max

### LINKED ISSUE #116: BUG: HBO Max not scrobbling + history sync not working - EU
HBO Max doesn't seem to work for me at all, neither scrobbling or sync. 

### When trying to sync history
If I have scrobble disabled and sync enabled in settings, a new tab for play.hbomax.com opens up and I just get stuck with the loading circle on the history page:
image

If I have scrobble and sync both enabled, a new tab quickly opens up and closes down and I get the error about not being logged in:
image

*I'm guessing the reason why I only get error popup when having scrobbling enabled is another bug.*

No error in Developer console and nothing in Developer network tab. when I'm trying to open history sync page. 

### When trying to scrobble
Nothing happens, no error in developer console and extension just shows the default "Watching something!"-window:
image

I would love to give more info to help solve this, just not sure how to collect relevant info?

rafaelgomesxyz:
Closes #116

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -45,6 +45,7 @@ const getJsOverride = () => {
 					hoist: 'all',
 				},
 			],
+			'prefer-named-capture-group': 'error',
 			'react/jsx-uses-react': 'off',
 			'react/react-in-jsx-scope': 'off',
 		},
--- a/src/apis/TraktSearch.ts
+++ b/src/apis/TraktSearch.ts
@@ -331,7 +331,7 @@ class _TraktSearch extends TraktApi {
 	formatEpisodeTitle(title: string): string {
 		return title
 			.toLowerCase()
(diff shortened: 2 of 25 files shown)


MrMamen (dismissed):
This seems to fix issues with History sync. However I still can't get scrobbling to work on movies. (episode scrobling works)

MrMamen:
It seems that the error is because the regexp creates two capture groups here:
https://github.com/trakt-tools/universal-trakt-scrobbler/blob/master/src/services/hbo-max/HboMaxParser.ts#L13
and it only support one here:
https://github.com/trakt-tools/universal-trakt-scrobbler/blob/master/src/common/ScrobbleParser.ts#L288

rafaelgomesxyz:
Got it, so it only happens with movies, right? Will add another commit here fixing it.

MrMamen (approved):

PR #139, opened 2022-03-14
Properly reset data before and after auto sync

rafaelgomesxyz:
Data is currently not properly reset before and after auto sync, which could lead to some leakage if the user navigates to the history after the auto sync, since they share the same store. This fixes that.


MrMamen (approved):
Nice! I have experienced this, but not able to pinpoint the error.

PR #146, opened 2022-03-16
Split Item and TraktItem models

rafaelgomesxyz:
`Item` and `TraktItem` are now split between `episode`, `show` and `movie`.

This PR has the potential to break a lot of things, because the structure of the models has been changed a bit. But I tested a bit and it seems to be working.

### DIFF
--- a/src/apis/TraktSearch.ts
+++ b/src/apis/TraktSearch.ts
@@ -4,7 +4,7 @@ import { CacheItems } from '@common/Cache';
 import { RequestError } from '@common/RequestError';
 import { Shared } from '@common/Shared';
 import { Utils } from '@common/Utils';
-import { EpisodeItem, isItem, ScrobbleItem } from '@models/Item';
+import { EpisodeItem, isItem, Item, ScrobbleItem } from '@models/Item';
 import {
 	createTraktScrobbleItem,
 	TraktEpisodeItem,
@@ -187,7 +187,7 @@ class _TraktSearch extends TraktApi {
 		}
 	}
 
-	async findItem(item: ScrobbleItem, cancelKey = 'default'): Promise<TraktSearchItem> {
+	async findItem(item: Item, cancelKey = 'default'): Promise<TraktSearchItem> {
 		let searchItem: TraktSearchItem | undefined;
 		await this.activate();
 		const responseText = await this.requests.send({
@@ -242,7 +242,7 @@ class _TraktSearch extends TraktApi {
 		if (!cacheItem) {
 			let show;
 			if (isItem(itemOrUrl)) {
-				show = ((await this.findItem(itemOrUrl, cancelKey)) as TraktSearchShowItem).show;
+				show = ((await this.findItem(itemOrUrl.show, cancelKey)) as TraktSearchShowItem).show;
 			} else {
 				await this.activate();
 				const showResponse = await this.requests.send({
--- a/src/apis/TraktSearch.ts
+++ b/src/apis/TraktSearch.ts
@@ -4,14 +4,20 @@ import { CacheItems } from '@common/Cache';
 import { RequestError } from '@common/RequestError';
 import { Shared } from '@common/Shared';
 import { Utils } from '@common/Utils';
-import { Item } from '@models/Item';
-import { TraktItem } from '@models/TraktItem';
+import { EpisodeItem, isItem, ScrobbleItem } from '@models/Item';
+import {
+	createTraktScrobbleItem,
+	TraktEpisodeItem,
+	TraktMovieItem,
+	TraktScrobbleItem,
+	TraktShowItemValues,
+} from '@models/TraktItem';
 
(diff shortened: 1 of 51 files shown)


src/models/Item.ts
```diff
@@ -27,94 +50,79 @@ export interface IItem extends ItemBase {
...
-	year?: number;
-	season?: number;
-	episode?: number;
-	episodeTitle?: string;
-	progress?: number;
+export type ShowItemParams = Omit<ShowItemValues, 'type' | 'trakt'> & {
+	trakt?: TraktShowItemParams | null;
+};
+
+export interface ShowItemValues extends BaseItemValues {
```
MrMamen, line 66: In terms of readability I would recommend switching order on some of these since Params are based on Values which are defined further down.

src/models/Item.ts
```diff
@@ -27,94 +50,79 @@ export interface IItem extends ItemBase {
...
+	type: 'show';
+	trakt?: TraktShowItemValues | null;
 }
 
-//TODO this should be refactored or split into show and movie. Inheritance could be used to get the similarities.
-export class Item implements IItem {
+export type MovieItemParams = Omit<MovieItemValues, 'type' | 'trakt'> & {
```
MrMamen: I guess this is to make it future proof, but as of now `MovieItemParams` and `ShowItemParams` could just as easily extend `BaseItemValues`, right?

src/models/TraktItem.ts
```diff
@@ -1,100 +1,202 @@
...
+	show: TraktShowItemValues;
+}
+
+export type TraktShowItemParams = Omit<TraktShowItemValues, 'type'>;
+
+export interface TraktShowItemValues extends TraktBaseItemValues {
+	type: 'show';
+}
+
+export type TraktMovieItemParams = Omit<TraktMovieItemValues, 'type'>;
```
MrMamen: Same here. It's hard to mentally track interfaces and types as they are defined in reversed order.

MrMamen (changes requested):
I really like the idea of splitting Item. However I found a few errors. The Episode/Season are sometimes returned as 0, and no match can be found. This used to work. See example screenshot below:
image
I'll continue my testing...

MrMamen:
Some of the error where because I forgot to migrate config.json to .env.
I haven't been able to pinpoint the exact problem, but it seems to search for episode titles even when season and episode information is available.

src/apis/TraktSearch.ts
```diff
@@ -220,27 +220,28 @@ class _TraktSearch extends TraktApi {
...
 		let traktDatabaseId = caches.urlsToTraktItems.get(showUrl);
-		let cacheItem = traktDatabaseId ? caches.traktItems.get(traktDatabaseId) : null;
+		let cacheItem = traktDatabaseId
+			? (caches.traktItems.get(traktDatabaseId) as TraktShowItemValues | undefined)
+			: null;
 		if (!cacheItem) {
 			let show;
-			if (itemOrUrl instanceof Item) {
+			if (isItem(itemOrUrl)) {
 				show = ((await this.findItem(itemOrUrl, cancelKey)) as TraktSearchShowItem).show;
```
MrMamen: It seems to never search for the show. In this location it actually searches for the episode, but it should search for the show. (I think it would be itemOrUrl.show)

src/modules/history/components/HistoryListItemCard.tsx
```diff
@@ -123,13 +123,13 @@ export const HistoryListItemCard = ({
 						<>
 							{item === null ? (
 								<Typography variant="h6">{I18N.translate('notFound')}</Typography>
-							) : item.type === 'show' ? (
+							) : item.type === 'episode' ? (
 								<>
-									{item.season && item.episode && (
-										<Typography variant="overline">{`S${item.season} E${item.episode}`}</Typography>
+									{item.season && item.number && (
```
MrMamen: If item.sesson == 0 , react outputs 0.

rafaelgomesxyz: Fixed

src/apis/TraktSearch.ts
rafaelgomesxyz: Fixed

src/models/TraktItem.ts
rafaelgomesxyz: Changed

src/models/Item.ts
rafaelgomesxyz: Changed

src/models/Item.ts
rafaelgomesxyz: Yeah, I think it's best to leave it like this for future changes, as I don't think it's a problem.

rafaelgomesxyz:
The updates I just pushed fixed the matching errors for me (remember to clear your cache when testing).

MrMamen (approved):
Seems to be working fine now!

PR #159, opened 2022-05-15
Create localization for French users, will be added to crowdin

tunisiano187:


rafaelgomesxyz:
Localizations should be added through an issue: https://github.com/trakt-tools/universal-trakt-scrobbler/issues/new?assignees=trakt-tools-bot&labels=new+language&template=new-language.md&title=Add+new+language%3A+%5BLANGUAGE%5D It will be automatically created on Crowdin and then you can add the translations there. We don't add translations through PRs.

PR #161, opened 2022-05-22
Netflix bug fixes and enhancements

### LINKED ISSUE #102: BUG: Incorrect scrobbling when using higher playback speed
When using a higher playback speed (ex. 2x - with a plugin), the scrobbling seems to be incorrect.
I noticed this on a Netflix show and on Prime.

I also get an error in console 'traktflix: Scrobble error undefined undefined undefined"
Not sure if this is due to the series: Outer banks s01e03 on Netflix -> shows "pilot" in traktflix

### LINKED ISSUE #153: Netflix - Not Identifying Season when name is Series
Hello,

I noticed that for some collections, the term Series is used instead of Season (See The Midwife - https://www.netflix.com/title/70245163). 

I think this has to do with the Netflix API (https://github.com/trakt-tools/universal-trakt-scrobbler/blob/master/src/services/netflix/NetflixApi.ts) on line 263 (const isCollection = !historyItem.seasonDescriptor.includes('Season');) is referencing a collection only by the static value "Season". I believe this needs to be updated to also include "Series" as part of the descriptor.

rafaelgomesxyz:
Bug fixes:

* Bug that scrobbles wrong episode (fixes #102)
* Bug that detects collections incorrectly (fixes #153)
* Bug that fails to match titles because of special characters

Enhancements:

* Fallback to first episode in response if unable to match by title

For testing:

[chrome.zip](https://github.com/trakt-tools/universal-trakt-scrobbler/files/8750199/chrome.zip)
[firefox.zip](https://github.com/trakt-tools/universal-trakt-scrobbler/files/8750200/firefox.zip)

### DIFF
--- a/src/apis/TraktSearch.ts
+++ b/src/apis/TraktSearch.ts
@@ -343,7 +343,7 @@ class _TraktSearch extends TraktApi {
 		return title
 			.toLowerCase()
 			.replace(/(?<begin>^|\s)(?:a|an|the)(?<end>\s)/g, '$<begin>$<end>')
-			.replace(/\s/g, '');
+			.replace(/[^\w]/g, '');
 	}
 }
 
--- a/src/apis/TraktSearch.ts
+++ b/src/apis/TraktSearch.ts
@@ -305,7 +305,7 @@ class _TraktSearch extends TraktApi {
 		showItem: TraktSearchShowItem,
 		episodeItems: TraktSearchEpisodeItem[]
 	): TraktSearchEpisodeItem {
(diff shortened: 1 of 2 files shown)


ITSYV:
I just tested it with an ridiculous fast speed and it seems to work at this moment.
I'll get back to this if it occurs again.

rafaelgomesxyz:
@MrMamen I believe this is ready for merging.

MrMamen (approved):
LGTM

PR #195, opened 2022-10-01
Fix sync storage on Firefox

rafaelgomesxyz:
Sync storage was broken for me when I just tested on Firefox. This fixes that.

### DIFF
--- a/src/common/BrowserStorage.ts
+++ b/src/common/BrowserStorage.ts
@@ -577,7 +577,7 @@ class _BrowserStorage {
 	 * and therefore can be saved in the sync storage.
 	 */
 	private async splitChunks(values: Record<string, unknown>): Promise<StorageValues> {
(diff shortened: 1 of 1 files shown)


src/common/BrowserStorage.ts
```diff
@@ -577,7 +577,7 @@ class _BrowserStorage {
 	 * and therefore can be saved in the sync storage.
 	 */
 	private async splitChunks(values: Record<string, unknown>): Promise<StorageValues> {
-		const maxSize = browser.storage.sync.QUOTA_BYTES_PER_ITEM;
+		const maxSize = browser.storage.sync.QUOTA_BYTES_PER_ITEM ?? 8192;
```
rafaelgomesxyz, line 580: 8192 is the default value on Chrome.

MrMamen (approved):

Issue #210, opened 2022-10-21
Crunchyroll sync not working


Catvari:
It's impossible to sync with crunchyroll. 
The extension always ask to login on the service even if it's already done.

![image](https://user-images.githubusercontent.com/55973375/197137040-a72c63fc-7f73-452a-b2a7-e11417bad1fa.png)
![image](https://user-images.githubusercontent.com/55973375/197137191-433dcbd1-9d8a-4130-b53a-59cd684b24fa.png)

noah978:
I believe this is due to the URL change. The new site is no longer using the "beta" subdomain. Its just normal crunchyroll now

cobrabr:
Any news on when #216 will be merged to `master`?

ipkpjersi:
I think this can be closed now? It seems like the MR was merged.

PR #216, opened 2022-11-06
Crunchyroll has left beta

### LINKED ISSUE #287: Crunchyroll no longer uses beta.crunchyroll.com (so it's broken on this scrobbler)
beta.crunchyroll.com now resolves to crunchyroll.com. It's been like this for a few months, and as such this tool has been broken for just as long. Could someone please change the code to reflect the correct URL so we can scrobble our anime lists?

MrMamen:
Removed beta from URLs, etc. However gets a CORS-error.


MrMamen:
Any idea @nyffchanium ?

nyffchanium:
@MrMamen I found the reason for the CORS error. The extension is not requesting access to the new "hostPatterns", thus the requests fail. After disabling and then re-enabling sync for crunchyroll, it worked just fine.
Not sure what to do about that though.

noah978:
Just noting that this solves #210 and looks good to me

MrMamen:
@rafaelgomesxyz Does this make sense?

src/common/BrowserStorage.ts
```diff
@@ -387,6 +391,18 @@ class _BrowserStorage {
 			console.log('Upgrading to v10...');
 		}
 
+		if (version < 11 && this.currentVersion >= 11) {
+			console.log('Upgrading to v11...');
+
+			const { options } = await this.get('options');
+
+			if (options?.services && 'crunchyroll-beta' in options.services) {
+				delete options.services['crunchyroll-beta'];
```
rafaelgomesxyz, line 400: We should copy it to `options.services['crunchyroll']` before deleting it, no? Otherwise the user options for Crunchyroll will be lost (they'll have to re-enable it).

src/common/BrowserStorage.ts
```diff
@@ -397,6 +413,16 @@ class _BrowserStorage {
 	 * They are only separated by type, to make it easier to understand the downgrade process.
 	 */
 	async downgrade(version: number) {
+		if (version > 10 && this.currentVersion <= 10) {
+			console.log('Downgrading to v10...');
+			const values = await this.get('options');
+			const options = values.options;
+			if (options) {
+				delete options.services['crunchyroll'];
```
rafaelgomesxyz, line 421: Same here. On the downgrade, we should copy it to `options.services['crunchyroll-beta']` before deleting it.

rafaelgomesxyz (changes requested):
Hey, sorry it took so long for me to get to this. Just minor comments.

src/common/BrowserStorage.ts
rafaelgomesxyz: I guess this ties back to https://github.com/trakt-tools/universal-trakt-scrobbler/pull/216#issuecomment-1356352686, so the users **should** re-enable it... In that case, disregard this and the comment below.

Still, it would be nice to warn people who have Crunchyroll enabled that they'd have to re-enable it, but that would be too much effort. I'm approving it.

rafaelgomesxyz (approved):

MrMamen:
The status of the linting is never reported. How come @rafaelgomesxyz ?

rafaelgomesxyz:
@MrMamen Oops, I renamed the action and forgot to update it in the PR merge requirements. Fixed it.

Issue #156, opened 2022-04-19
Update to manifest v3


MrMamen:
Manifest v2 is deprecated and no longer accepted for releases:
https://developer.chrome.com/blog/mv2-transition/

This rewrite is required for #147

rafaelgomesxyz:
This will be a bit of a problem because of how we use the background page at the moment, but I'll try to figure something out.

rafaelgomesxyz:
Will try to tackle this. From an initial observation, this is what needs to be done:

- [x] General changes to the manifest file. These will probably not be that difficult, as it's mostly renaming things and moving things around.
- [ ] Adapt any logic that required the background page to stay active (service workers go inactive after a while). Most of this is probably achievable by storing the data with `browser.storage` and then getting it when needed. Major refactor.
- [x] Adapt script injections to use `scripting.executeScript`. Small refactor.
- [x] Adapt any code that ran in intervals to use `browser.alarms` (like auto sync and cache checks running every hour). Small refactor.
- [ ] Adapt usages of `webRequest` to the new `declarativeNetRequest`. This one might be the more annoying one, but since `webRequest` is only used for Firefox, and only when using containers, I might just ignore it for now.

PR #218, opened 2022-11-20
Replace setTimeout with alarms

rafaelgomesxyz:
This begins the effort of making the code compatible with Manifest v3 (https://github.com/trakt-tools/universal-trakt-scrobbler/issues/156), so it's easier to upgrade to it.

`alarms` doesn't require the background page to keep running, so it solves the issue of having to trigger cache / auto sync checks every 60 minutes.

### DIFF
--- a/src/common/AutoSync.ts
+++ b/src/common/AutoSync.ts
@@ -42,8 +42,6 @@ class _AutoSync {
 	private async check() {
 		await Shared.waitForInit();
 
-		console.log('Checking auto sync...');
-
 		const now = Utils.unix();
 		const servicesToSync = getServices().filter((service) => {
 			const value = Shared.storage.options.services[service.id];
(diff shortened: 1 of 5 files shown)


MrMamen (approved):

src/common/AutoSync.ts
```diff
@@ -12,37 +11,38 @@ import { getServices, Service } from '@models/Service';
...
-		if (this.checkTimeoutId !== null) {
-			window.clearTimeout(this.checkTimeoutId);
-		}
+		void this.check();
+	};
+
+	private async check() {
+		await Shared.waitForInit();
+
+		console.log('Checking auto sync...');
```
MrMamen: Do you need this in prod, or was it just for debugging?

rafaelgomesxyz: Debugging, will remove.

Issue #242, opened 2023-01-02
Hide trailers?


FezVrasta:
On Prime Video I watched "The Grand Tour: A Scandi Flick - Trailer" and it shows up as a real movie rather than a trailer. Do you think it would make sense to hide elements that have "trailer" in the name?

downey-lv:
In Prime API there are additional entityType:
-"Bonus Content" 
-"Trailer"

We should probably ignore (is it possible to ignore items?) Trailer and try to match Bonus Content against Trakt specials. I'll look into it

UPDATE: Trailers are often also marked as Bonus Content, so it isn't as straightforward

downey-lv:
> Do you think it would make sense to hide elements that have "trailer" in the name?

This wouldn't be a good approach due to mismatches in both directions, e.g. "Trailer Park Boys" as false positives and "The Grand Tour presents… Carnage A Trois" (same name for trailer and episode) as false negatives.

PR #249, opened 2023-01-06
Add support for Prime bonus content

downey-lv:
Added Bonus Content to episode logic, this allows to sync Prime Bonus content with Specials on Trakt. 

Previously bonus content was handled as movie by default, which led to no matches found or incorrect matches (while testing I saw a Rammstein concert as a match for one of the bonus videos).

Found the two new entityTypes after looking into #242.

Before:
image

After:
image


MrMamen (approved):
Thanks for your continuous contributions.

PR #253, opened 2023-01-13
Fixes history in Firefox containers + some small additional changes

### LINKED ISSUE #252: Syncing history with Firefox containers is impossible
Currently, it is impossible to open the history sync page when using Firefox containers, as it always says, that you are not logged in.

chaptergy:
This mainly fixes https://github.com/trakt-tools/universal-trakt-scrobbler/issues/252.  
The issue occurred, because the regex pattern to get the domain part of the URL actually got the path.

In addition, it has some other minor improvements:

- Added a comment to the readme in the development section to also add an extension ID, because otherwise the messaging is unavailable.
- Added fallback in messaging error handling in case of the error not actually being a JSON string. It will now return the original error, which would otherwise be hard to get and debug.
- Only sets the `isActivated` flag for Netflix if it has a profile name. Otherwise, you are stuck with an invalid session, while the `activate()` method will no longer be called. The session object is still set, even when Netflix is not logged in, hence only the profile name can help check, whether the user is _actually_ logged in

### DIFF
--- a/src/common/Messaging.ts
+++ b/src/common/Messaging.ts
@@ -241,13 +241,17 @@ class _Messaging {
 			response = ((await browser.runtime.sendMessage(message)) ?? null) as ReturnType<T>;
 		} catch (err) {
 			if (err instanceof Error) {
-				const messagingError = JSON.parse(err.message) as MessagingError;
-				switch (messagingError.instance) {
-					case 'Error':
-						throw new Error(messagingError.data.message);
+				try {
+					const messagingError = JSON.parse(err.message) as MessagingError;
+					switch (messagingError.instance) {
+						case 'Error':
+							throw new Error(messagingError.data.message);
 
-					case 'RequestError':
-						throw new RequestError(messagingError.data);
+						case 'RequestError':
+							throw new RequestError(messagingError.data);
+					}
+				} catch (_) {
+					throw err;
 				}
 			}
 		}
--- a/src/common/Requests.ts
+++ b/src/common/Requests.ts
@@ -150,7 +150,7 @@ class _Requests {
 		if (!Shared.storage.options.grantCookies || !browser.cookies || !browser.webRequest) {
 			return;
 		}
-		const domainMatches = /https?:\/\/(?:www\.)?(?:.+?)(?<domain>\/.*)?$/.exec(request.url);
+		const domainMatches = /https?:\/\/(?:www\.)?(?<domain>.+?)(?:\/.*)?$/.exec(request.url);
 		if (!domainMatches?.groups) {
 			return;
 		}
--- a/src/services/netflix/NetflixApi.ts
+++ b/src/services/netflix/NetflixApi.ts
@@ -188,7 +188,7 @@ class _NetflixApi extends ServiceApi {
 				});
 				this.session = this.extractSession(responseText);
 			}
-			if (this.session && this.session.profileName) {
+			if (this.session?.profileName != null) {
 				this.isActivated = true;
 			}
 		} catch (err) {
@@ -203,7 +203,7 @@ class _NetflixApi extends ServiceApi {
 		if (!this.isActivated) {
 			await this.activate();
 		}
-		return !!this.session && this.session.profileName !== null;
+		return this.session?.profileName != null;
 	}
 
 	async loadHistoryItems(): Promise<NetflixHistoryItem[]> {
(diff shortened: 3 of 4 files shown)


src/common/Messaging.ts
```diff
@@ -241,13 +241,17 @@ class _Messaging {
...
 		} catch (err) {
 			if (err instanceof Error) {
-				const messagingError = JSON.parse(err.message) as MessagingError;
-				switch (messagingError.instance) {
-					case 'Error':
-						throw new Error(messagingError.data.message);
-
-					case 'RequestError':
-						throw new RequestError(messagingError.data);
+				try {
```
rafaelgomesxyz, line 244: It doesn't feel right for me to have a `try...catch` inside of a `try...catch`, but it does fix the issue, so maybe we can leave it like this for now.

src/common/Requests.ts
```diff
@@ -150,7 +150,7 @@ class _Requests {
 		if (!Shared.storage.options.grantCookies || !browser.cookies || !browser.webRequest) {
 			return;
 		}
-		const domainMatches = /https?:\/\/(?:www\.)?(?:.+?)(?<domain>\/.*)?$/.exec(request.url);
+		const domainMatches = /https?:\/\/(?:www\.)?(?<domain>.+?)(?:\/.*)?$/.exec(request.url);
```
rafaelgomesxyz, line 153: Thanks! I actually found and fixed this issue as well on my local, but haven't had time to push it yet.

src/services/netflix/NetflixApi.ts
```diff
@@ -188,7 +188,7 @@ class _NetflixApi extends ServiceApi {
 				});
 				this.session = this.extractSession(responseText);
 			}
-			if (this.session) {
+			if (this.session && this.session.profileName) {
```
rafaelgomesxyz: Why not `this.session?.profileName`?

src/common/Messaging.ts
chaptergy: Yeah, I tried finding an alternative option, but I couldn't think of a different quick fix...

src/services/netflix/NetflixApi.ts
chaptergy: Sure, that optimizes the code a bit, I just used a similar style as in the [`checkLogin()`](https://github.com/trakt-tools/universal-trakt-scrobbler/blob/14efdd20bccc0815f5cc5cad5063cd2b16145024/src/services/netflix/NetflixApi.ts#L206)-function. But I'll go ahead and update it.

src/services/netflix/NetflixApi.ts
```diff
@@ -188,7 +188,7 @@ class _NetflixApi extends ServiceApi {
 				});
 				this.session = this.extractSession(responseText);
 			}
-			if (this.session && this.session.profileName) {
+			if (this.session?.profileName != null) {
```
rafaelgomesxyz, line 191: I don't think we need to compare against `null` here. Just `this.session?.profileName` should work, as it would only be `true` if `profileName` is a non-empty string.

src/services/netflix/NetflixApi.ts
```diff
@@ -203,7 +203,7 @@ class _NetflixApi extends ServiceApi {
 		if (!this.isActivated) {
 			await this.activate();
 		}
-		return !!this.session && this.session.profileName !== null;
+		return this.session?.profileName != null;
```
rafaelgomesxyz, line 206: Same here, this could be changed to `!!this.session?.profileName`.

src/services/netflix/NetflixApi.ts
chaptergy: I'm not sure, whether it is possible to have a profile with an empty string as a name. I chose to make the not nullish comparison, because I'm pretty sure when you are not logged in, `profileName` is `null`, and when the session is not set, it is either `null` or `undefined`, which are both covered by this comparison.

But of course if you prefer, I can change it to a truthy comparison.

rafaelgomesxyz: > I'm not sure, whether it is possible to have a profile with an empty string as a name.

I don't think so, it should be safe to change it.

rafaelgomesxyz: Actually, it might be the case that we don't have access to the profile name... Hm, maybe we can leave it like this then.

rafaelgomesxyz (approved):

PR #255, opened 2023-01-15
Updates the name of cli from trakt-tools to @trakt-tools/cli

chaptergy:
Updates the README to contain the correct CLI package name

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,7 +53,7 @@ If you want to scrobble / sync from Netflix, this is the only Trakt.tv [plugin](
 ### Which streaming services are supported?
 
 <!-- services-start -->
-<!-- Update this section with `npx trakt-tools dev update-readme` -->
+<!-- Update this section with `npx @trakt-tools/cli dev update-readme` -->
 
 | Streaming Service | Scrobble | Sync | Limitations                     |
 | :---------------: | :------: | :--: | :------------------------------ |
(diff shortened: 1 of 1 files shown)


rafaelgomesxyz:
Just `trakt-tools` should work, because of https://github.com/trakt-tools/cli/blob/main/package.json#L8. Did it not work for you?

chaptergy:
> Just `trakt-tools` should work, because of https://github.com/trakt-tools/cli/blob/main/package.json#L8. Did it not work for you?

Using `trakt-tools` directly is only possible, if `@trakt-tools/cli` is already installed. Otherwise you get the following error:

```
> npx trakt-tools
npm ERR! code E404
npm ERR! 404 Not Found - GET https://registry.npmjs.org/trakt-tools - Not found
npm ERR! 404 
npm ERR! 404  'trakt-tools@latest' is not in this registry.
npm ERR! 404 
npm ERR! 404 Note that you can also install from a
npm ERR! 404 tarball, folder, http url, or git url.
```

rafaelgomesxyz:
That's the thing, it should always be installed if you installed the dependencies (which should be the first step after cloning the repo anyway), because it's listed in the dev dependencies.

It's a lot faster to type `npx trakt-tools` than `npx @trakt-tools/cli`, so that's why I'd prefer not to do this change if we don't need to.

chaptergy:
You're right, I'm sorry. I forgot, that when I encountered the issue I reset my git folder right before, causing the node modules to be lost

PR #256, opened 2023-01-15
Make compatible with Manifest v3

### LINKED ISSUE #156: Update to manifest v3

rafaelgomesxyz:
Fixes #156

**NOTE:** This should be carefully tested, as some things might still be broken. I'll continue to test on my end throughout this week.

This follows Chrome's [migration guide](https://developer.chrome.com/docs/extensions/mv3/mv3-migration/) and:

- [x] changes manifest version to v3
- [x] changes background script to service worker
- [x] separates permissions from host permissions
- [x] removes content security policy
- [x] renames browser action to action
- [x] removes web accessible resources
- [x] uses scripting API to inject content scripts
- [x] uses scripting API to inject functions (all functions that need to be injected now have to be added to `Shared.functionsToInject` on first run, because they need to be accessible from the service worker)
- [x] removes ports, as they're not compatible with service workers (the new method to identify when a content script has connected is to send a message to the service worker, and the new method to identify when a content script has disconnected is to add a listener to `browser.tabs.onRemoved`)
- [x] removes axios, as `XMLHttpRequest` isn't compatible with service workers (rate limiting has been removed as well, but our requests are already sequential, so they shouldn't cause many issues)
- [x] removes window/document references from background logic, as they're not available in service workers

There might be other changes that need to be made.

These changes should be fully backwards compatible with manifest v2. Webpack is configured to build v2 for Firefox and v3 for Chrome, as v3 isn't accepted in the Firefox store yet, as far as I could tell.

### DIFF
--- a/src/services/netflix/NetflixApi.ts
+++ b/src/services/netflix/NetflixApi.ts
@@ -445,19 +445,7 @@ class _NetflixApi extends ServiceApi {
 	}
 
 	getSession(): Promise<NetflixSession | null> {
-		return ScriptInjector.inject<NetflixSession>(this.id, 'session', '', () => {
-			let session: NetflixSession | null = null;
-			const { netflix } = window;
-			if (netflix) {
-				const { userInfo } = netflix.reactContext.models;
-				const authUrl = userInfo.data.authURL;
-				const profileName = userInfo.data.name;
-				if (authUrl) {
-					session = { authUrl, profileName };
-				}
-			}
-			return session;
-		});
+		return ScriptInjector.inject<NetflixSession>(this.id, 'session', '');
 	}
 
 	extractSession(text: string): NetflixSession | null {
@@ -472,5 +460,19 @@ class _NetflixApi extends ServiceApi {
 		return session;
 	}
 }
+
+Shared.functionsToInject[`${NetflixService.id}-session`] = () => {
+	let session: NetflixSession | null = null;
+	const { netflix } = window;
+	if (netflix) {
+		const { userInfo } = netflix.reactContext.models;
+		const authUrl = userInfo.data.authURL;
+		const profileName = userInfo.data.name;
+		if (authUrl) {
+			session = { authUrl, profileName };
+		}
+	}
+	return session;
+};
 
 export const NetflixApi = new _NetflixApi();
--- a/src/services/netflix/NetflixApi.ts
+++ b/src/services/netflix/NetflixApi.ts
@@ -206,7 +206,7 @@ class _NetflixApi extends ServiceApi {
 		return this.session?.profileName != null;
 	}
 
-	async loadHistoryItems(): Promise<NetflixHistoryItem[]> {
+	async loadHistoryItems(cancelKey = 'default'): Promise<NetflixHistoryItem[]> {
 		if (!this.isActivated) {
 			await this.activate();
 		}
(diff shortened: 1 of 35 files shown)


rafaelgomesxyz:
@MrMamen It's a WIP, but if you already want to take a look.

rafaelgomesxyz:
@MrMamen Can we try to land this? I didn't test too much (also mostly just Netflix), but we can always refine it later. I'd like to release it on Chrome soon.

MrMamen:
Yes. I'll test it a bit.

src/services/nrk/NrkParser.ts
```diff
@@ -1,21 +1,23 @@
...
 
-	itemIdFnToInject = () => {
-		let itemId: string | null = null;
-		const { player } = window;
-		if (player) {
-			const playbackSession = player.getPlaybackSession();
-			if (playbackSession) {
-				itemId = playbackSession.mediaItem?.id ?? null;
-			}
+Shared.functionsToInject[`${NrkApi.id}-item-id`] = () => {
```
MrMamen: Not sure why, but this code is never run, thus scrobbling fails

MrMamen:
NRK doesn't work with scrobbeling. This is due to the injection script not working. (Only tested on Chrome).

rafaelgomesxyz:
Got it, thanks, I'll see if I can fix it.

rafaelgomesxyz:
@MrMamen Script injection for functions should be fixed.

MrMamen:
> @MrMamen Script injection for functions should be fixed.

It seems to me that the scrobbling part needed to be disabled and re-enabled after switching to this branch. (for NRK at least.) If there is no way to prevent this, perhaps the migrate script should simply disable the scrobbling for injection-script sites?

rafaelgomesxyz:
> > @MrMamen Script injection for functions should be fixed.
> 
> It seems to me that the scrobbling part needed to be disabled and re-enabled after switching to this branch. (for NRK at least.) If there is no way to prevent this, perhaps the migrate script should simply disable the scrobbling for injection-script sites?

But that's for Chrome, right? I don't think it will be possible to migrate from v2 to v3 on Chrome, because they have different IDs, so they should be two separate extensions when installed. That wasn't the case for you because you probably don't have the ID configured in your `.env`. Unfortunately, I think Chrome users will have to reconfigure their options.

rafaelgomesxyz:
I'll see if I can add an import/export functionality to allow users to move their data.

MrMamen:
> I'll see if I can add an import/export functionality to allow users to move their data.

Om not sure how important this is. More important to get this to shipped. We should be more careful with these things after it hits the Web Stores.

rafaelgomesxyz:
Ok, I've actually already uploaded it to the Chrome store because I wanted to test something: https://chrome.google.com/webstore/detail/universal-trakt-scrobbler/mbhadeogepkjdjeikcckdkjdjhhkhlid

But before we add it to the `README` I want to land another PR that I'll open later today that allows users to do a full initial sync.

MrMamen:
I experience that at least netflix history keeps loading/spinning forever, with no obvious error..

src/services/netflix/NetflixApi.ts
```diff
@@ -444,19 +444,7 @@ class _NetflixApi extends ServiceApi {
...
-				const { userInfo } = netflix.reactContext.models;
-				const authUrl = userInfo.data.authURL;
-				const profileName = userInfo.data.name;
-				if (authUrl) {
-					session = { authUrl, profileName };
-				}
-			}
-			return session;
-		});
+		return ScriptInjector.inject<NetflixSession>(this.id, 'session', '');
```
MrMamen, line 448: I've pinpointed that this is the point of failure.

rafaelgomesxyz: @MrMamen I've pushed some changes, can you see if it's fixed now?

MrMamen (approved):
Looks good to me.

PR #259, opened 2023-01-15
Improved error logging in some places

chaptergy:
Shows actual request status when request fails and logs error when calling non-implemented ServiceApi function.

I added this, because I was stuck in these places for a while before noticing, that something went wrong.


MrMamen (approved):

MrMamen:
Thanks

PR #261, opened 2023-01-17
Improves history view

chaptergy:
This PR has various changes, which (in my opinion) improve mainly the history view a bit. These are:

- Episode titles are ellipsized, when they are too long, but movie titles and show names were not. This unifies the behavior, so all titles are ellipsized.
- When initially loading the history list (not from cache) and reaching the end of the list, some items at the very end could be missing. This happened, when the number of items in the history is not divisible by 10, but the API returns more than 10 history items at a time. A fix is included, so these items are shown as well.
- Adds an "unknown" sync status, which is used, as long as there is no associated trakt item. This prevents the sync status from being shown as red / not synced while the history is being loaded and represents the actual internal state a bit better. Instead, it will now show a gray arrow (instead of a red or a green one) while loading.
- Includes information about previous watches for unsynced watches. Until now, it said "not watched" if the watchedAt value could not be associated with a trakt item with a similar watchedAt value, which is technically not correct. Now it says "watched x other times" to let the user know, that they already have an entry in their trakt history, but it was at other points in time. Hovering over this text also shows at which dates they have watchet it.
- Currently, an image provided by the serviceApi will be shown behind the trakt item. This can cause confusion, because if the trakt item is not actually mapped correctly it might seem like it is, because the background image is correct. In this PR this behavior is changed, so images provided by the serviceApi are shown as the background of the service item card and the background is always an image from TMDB for the actual trakt item.
- When disabling the "load images" setting, actually no images are now loaded in the history view, instead of not just collecting the TMDB URLs. Image ULRs provided by the serviceApi would have been loaded anyway.
- Changed / removed some fixed colors to be based on the theme (this is in other parts as well, not just the history view). Mainly, the bottom bar in the settings and history view was set to always be white, which is now changed to be the theme's background color.

### DIFF
--- a/src/_locales/en/messages.json
+++ b/src/_locales/en/messages.json
@@ -316,6 +316,9 @@
 	"options": {
 		"message": "Options"
 	},
+	"otherPlays": {
+		"message": "Other plays:"
+	},
 	"profile": {
 		"message": "Profile"
 	},
@@ -451,6 +454,14 @@
 	},
 	"watched": {
 		"message": "Watched On"
+	},
+	"watchedOtherTimes": {
+		"message": "Watched $COUNT$ other times",
+		"placeholders": {
+			"count": {
+				"content": "$1"
+			}
+		}
 	},
 	"yes": {
 		"message": "Yes"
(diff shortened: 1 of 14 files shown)


MrMamen (approved):
Great improvements. Sorry for taking so long to test this.

PR #286, opened 2023-02-26
Universal Amazon Prime endpoints

MrMamen:
- Fix prime sync for Prime Brazil
- Universal Region fix for Amazon Prime
Co-authored-by: Throne3d <Throne3d@users.noreply.github.com>

PR #301, opened 2023-04-12
use alternative api endpoints when www.amazon.* is used as HOST_URL

Superritchman:
[${CONFIG_URL}](https://github.com/trakt-tools/universal-trakt-scrobbler/blob/3b2ab19a23a9f322b2911fd7a0f4d2be8a10cbe9/src/services/amazon-prime/AmazonPrimeApi.ts#L159) endpoint could return an amazon url as `defaultVideoWebsite` e.g. https://www.amazon.de

In #286 the endpoints changed in favor of the regional primevideo host url.
But with a host url like https://www.amazon.de, the api endpoints are not valid.


rafaelgomesxyz (approved):
Since my `HOST_URL` is `primevideo.com`, I'll just take you at your word that it works for you. Thanks for the contribution!

PR #308, opened 2023-05-06
Support scrobbling in TV 2 Play

MrMamen:
I have an almost working history sync as well. But I can't seem to find a way to retrieve the token (JWT) for auth.

### DIFF
--- a/src/services/tv2-play/Tv2PlayApi.ts
+++ b/src/services/tv2-play/Tv2PlayApi.ts
@@ -25,8 +25,6 @@ interface TV2SumoMetadataResponseMovie {
 }
 
 class _Tv2PlayApi extends ServiceApi {
-	// Define any properties you need here
-
 	constructor() {
 		super(Tv2PlayService.id);
 	}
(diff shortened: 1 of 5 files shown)


src/services/tv2-play/Tv2PlayApi.ts
```diff
@@ -0,0 +1,66 @@
...
+}
+
+interface TV2SumoMetadataResponseMovie {
+	id: number;
+	title: string;
+	asset_type: 'movie';
+}
+
+class _Tv2PlayApi extends ServiceApi {
+	// Define any properties you need here
```
rafaelgomesxyz: Not needed.

src/services/tv2-play/Tv2PlayApi.ts
```diff
@@ -0,0 +1,66 @@
...
+				},
+			};
+			return new EpisodeItem(values);
+		} else {
+			const values = {
+				serviceId: this.id,
+				id,
+				title: responseJson.title,
+			};
+			return new MovieItem(values);
```
rafaelgomesxyz: No need to put this code in an `else` block, since there's a `return` in the `if` block.

rafaelgomesxyz (approved):
LGTM. Just a couple minor things if you can change them before merging, but not a big deal.

PR #309, opened 2023-05-15
Hotstar: update selector to parse data

fikimaul:
Hotstar have major update their website and have different selector then before. 
So, This pull request will make scroble working again.

Movie

movie

Shows
show

### DIFF
--- a/src/services/hotstar/HotstarParser.ts
+++ b/src/services/hotstar/HotstarParser.ts
@@ -5,7 +5,7 @@ import { EpisodeItem, MovieItem } from '@models/Item';
 class _HotstarParser extends ScrobbleParser {
 	constructor() {
 		super(HotstarApi, {
-			watchingUrlRegex: /\/(?:movies|shows)\/\D+(?<id>\d.+)\//,
+			watchingUrlRegex: /(?:movies|shows)\/.+\/(?<id>\d+)\//,
 		});
 	}
 
--- a/src/services/hotstar/HotstarParser.ts
+++ b/src/services/hotstar/HotstarParser.ts
@@ -5,35 +5,36 @@ import { EpisodeItem, MovieItem } from '@models/Item';
 class _HotstarParser extends ScrobbleParser {
 	constructor() {
 		super(HotstarApi, {
-			watchingUrlRegex: /\/(?:movies|tv)\/(?<id>.+)/,
+			watchingUrlRegex: /\/(?:movies|shows)\/\D+(?<id>\d.+)\//,
 		});
 	}
 
 	parseItemFromDom() {
 		const serviceId = this.api.id;
 		const id = this.parseItemIdFromUrl();
-		const titleElement = document.querySelector('.primary-title');
+		const titleElement = document.querySelector(
+			'div#page-container>div>div>div>div>div>div>div>div>div>div.flex.flex-col>div.flex>p.ON_IMAGE.BUTTON1_MEDIUM'
+		);
 
 		if (!titleElement) {
 			return null;
 		}
 
 		const title = titleElement?.textContent ?? '';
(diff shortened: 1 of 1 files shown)


src/services/hotstar/HotstarParser.ts
```diff
@@ -5,35 +5,36 @@ import { EpisodeItem, MovieItem } from '@models/Item';
...
+			watchingUrlRegex: /\/(?:movies|shows)\/\D+(?<id>\d.+)\//,
 		});
 	}
 
 	parseItemFromDom() {
 		const serviceId = this.api.id;
 		const id = this.parseItemIdFromUrl();
-		const titleElement = document.querySelector('.primary-title');
+		const titleElement = document.querySelector(
+			'div#page-container>div>div>div>div>div>div>div>div>div>div.flex.flex-col>div.flex>p.ON_IMAGE.BUTTON1_MEDIUM'
```
rafaelgomesxyz, line 16: Can we use a more specific selector?

src/services/hotstar/HotstarParser.ts
```diff
@@ -5,35 +5,36 @@ import { EpisodeItem, MovieItem } from '@models/Item';
...
 			return null;
 		}
 
 		const title = titleElement?.textContent ?? '';
-		const seasonEpisodeElement = document.querySelector(
-			'.show-title .meta-data-holder'
-		)?.firstChild;
-		const subTitleElement = document.querySelector('.show-title .meta-data-holder')?.lastChild;
+		const subTitleElement = document.querySelector(
+			'div#page-container>div>div>div>div>div>div>div>div>div>div.flex.flex-col>p.ON_IMAGE_ALT2.BUTTON3_MEDIUM'
```
rafaelgomesxyz, line 25: Can we use a more specific selector?

src/services/hotstar/HotstarParser.ts
```diff
@@ -5,35 +5,36 @@ import { EpisodeItem, MovieItem } from '@models/Item';
 class _HotstarParser extends ScrobbleParser {
 	constructor() {
 		super(HotstarApi, {
-			watchingUrlRegex: /\/(?:movies|tv)\/(?<id>.+)/,
+			watchingUrlRegex: /\/(?:movies|shows)\/\D+(?<id>\d.+)\//,
```
rafaelgomesxyz: I don't think this is correct for episodes.

Using the URLs from your screenshots:

- `hotstar.com/id/shows/dr-stone/1260117151/dr-stone/1260117832/watch?filters=content_type%3Depisode` results in `1260117151/dr-stone/1260117832` 
- `hotstar.com/id/movies/sri-asih/1260133861/watch?fiIters=content_type%3Dmovie` results in `1260133861`

I think for episodes it should result in `1260117832`.

So a better regex would probably be: `/(?:movies|shows)\/.+\/(?<id>\d+)\//`

rafaelgomesxyz (changes requested):

src/services/hotstar/HotstarParser.ts
fikimaul: I think chaining selector is best way to get the element. Because in the new website no more id/class in specific. When i've only use `p.ON_IMAGE.BUTTON1_MEDIUM` selector, will get content rating that show and disapear in a few seconds. And when use class like `_2Xo-ONKnvHxYr5Qupt1eXf`, i think will change regulary when site has new update because the class is generated by library they use. 

image

src/services/hotstar/HotstarParser.ts
fikimaul: Thank you for correction. The regex was updated

rafaelgomesxyz (approved):

PR #319, opened 2023-07-01
Add initial sync

### LINKED ISSUE #47: Give user option to do initial sync and prevent loading synced history

rafaelgomesxyz:
Fixes #47

I tested on Netflix and HBO Max and it seems to work well.

### DIFF
--- a/src/_locales/en/messages.json
+++ b/src/_locales/en/messages.json
@@ -204,6 +204,18 @@
 	"home": {
 		"message": "Home"
 	},
+	"initialSyncTitle": {
+		"message": "Sync full history?"
+	},
+	"initialSyncDescription": {
+		"message": "This is the first time that you are accessing this service. Would you like to sync the full history right now? This may take a while depending on how big your history is.\n\nBefore choosing \"Yes\", make sure that the options below are adjusted to your liking.\n\nIf you choose \"No\", you will not be asked this again, unless you clear the last sync date for the service or clear the storage."
+	},
+	"initialSyncSuccess": {
+		"message": "Sync finished with success. Would you like to review it and make sure that the correct items were synced?"
+	},
+	"initialSyncFailed": {
+		"message": "Sync failed. Would you like to review it and correct the items?"
+	},
 	"invalidTraktUrl": {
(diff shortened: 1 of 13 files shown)


rafaelgomesxyz:
I actually noticed a bug with the logic to retry requests after 429 status, will fix and add commits here.

MrMamen:
> I actually noticed a bug with the logic to retry requests after 429 status, will fix and add commits here.

Let me know when it is ready for review....

src/_locales/en/messages.json
```diff
@@ -204,6 +204,18 @@
 	"home": {
 		"message": "Home"
 	},
+	"initialSyncTitle": {
+		"message": "Sync full history?"
+	},
+	"initialSyncDescription": {
+		"message": "This is the first time that you are accessing this service. Would you like to sync the full history right now? This may take a while depending on how big your history is.\n\nBefore choosing \"Yes\", make sure that the options below are adjusted to your liking.\n\nIf you choose \"No\", you will not be asked this again, unless you clear the last sync date for the service or clear the storage."
```
jerone, line 211:
```suggestion
		"message": "This is the first time that you are accessing this service. Would you like to sync the full history right now? This may take a while depending on how big your history is.\n\nBefore choosing \"Yes\", make sure that the options below are adjusted to your liking.\n\nIf you choose \"No\", you will not be asked this again, unless you clear the last sync date for the service or clear the storage. You can always trigger the sync manually."
```
Makes it less of a now-or-never condition.

Issue #342, opened 2023-09-21
Prime Video History - ignore [dt./OV] tags in Name


Poddi:
Trakt wont find these Items.
![Screenshot 2023-09-22 000730](https://github.com/trakt-tools/universal-trakt-scrobbler/assets/51752434/6c8ed0bd-9fa7-4b0e-aada-489acb522202)

smoothlystable:
Is it this movie? https://trakt.tv/movies/wer-2013

Easiest would be to just click "Ist das falsch?" and paste the link. 
If I'm looking at the correct movie, the "[dt./OV]" is part of the movie name on Prime, so it probably shouldn't be ignored.

Poddi:
Its this Movie : https://trakt.tv/movies/synchronic-2020

This also happens on every Episode for Supernatural : https://trakt.tv/shows/supernatural
And many others.
![image](https://github.com/trakt-tools/universal-trakt-scrobbler/assets/51752434/8f389ac1-81b1-458c-8152-fd85e447bf81)

I started to paste the Links manually but then i noticed its too much.

smoothlystable:
Ooh, never mind then. I've never seen it before, so I got confused, I can't reproduce it with Supernatural either;
![image](https://github.com/trakt-tools/universal-trakt-scrobbler/assets/20187834/7fcc9bf8-3081-4440-87dd-e3bb8a1aeb34)

I'm guessing "[dt./OV]" has something to do with watching a special version of the movie/show? Are you watching a German version perhaps (like dubbed or something)?

The-Kiyoshi:
i have the same problem and yes i watched it in german @smoothlystable 
<img width="851" alt="image" src="https://github.com/trakt-tools/universal-trakt-scrobbler/assets/57962401/baa03a5a-6ce7-4cc1-bd6f-bfa4d52cac91">

The-Kiyoshi:
isnt the easiest way to just look is there [dt./OV] if yes remove? @smoothlystable

Poddi:
Hello,
its working for some but not all.
Can you please check again .
Here are some Examples which are still not working.
![Screenshot 2023-10-07 124724](https://github.com/trakt-tools/universal-trakt-scrobbler/assets/51752434/b69ec9e2-592c-467a-bce9-0a65916c3e27)
[...]

PR #344, opened 2023-10-05
AmazonPrime: Remove [dt./OV] from movie/show title

smoothlystable:
Seems like when playing special versions of a show/movie, Prime adds a tag to the title, this messes with the Trakt search. 

German versions seems to add "[dt./OV]", not sure if other tags exists.

Closes #342

PR #346, opened 2023-10-15
AmazonPrime: Remove all version tags from movie/show title

smoothlystable:
This is a follow up on the #344 pull request, since there seems to be a lot more tags than just one, I changed it to regex instead. 

Hopefully this will close issue #342 once and for all.


MrMamen:
Sorry. Reusing the same branch causes problem with merging due to merge conflict. Would you mind resolve the conflicts or create a new based on the current master branch?

smoothlystable:
Sorry about that, should be fixed now.

MrMamen (approved):

PR #348, opened 2023-11-11
Implement support for Crave

g0tsimid:
Add scrobbling and watch history sync support for https://www.crave.ca (Canadian streaming service).

### Details

- **Service Name:** Crave
- **Service URL:** https://www.crave.ca
- **Is it region-locked? If yes, to what region?** Yes, Canada
- **Does it have a free version / trial?** No

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,13 +39,19 @@ Extension will be enabled until you restart Firefox.
 
 ### Table of Contents
 
+- [Loading the extension manually in Chrome](#loading-the-extension-manually-in-chrome)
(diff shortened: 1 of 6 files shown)


MrMamen (approved):
Thanks for your contribution, and sorry for the very late review!
I've not tested the integration but the code looks good. Nice to have a fallback way of parsing items if the API is stale.

MrMamen:
Why aren't the github checks running? How can I trigger them @rafaelgomesxyz ?

PR #355, opened 2023-11-21
add support for Max based on HBO Max setup

varnent:


dan-riley:
Do these changes actually work?  Don't want to waste time trying to compile it if not.  I tried changing all of the url's to max.com in the unpacked install, but it didn't work.  Quick look at this commit looks like that's all it does too essentially, but I might be missing something.

MrMamen:
Can someone in a region where Max is available test and verify this?

s-crypt:
Idk if I am doing something wrong, but I cannot get the Max toggle to turn on

jarrodsmeyers:
Even after fixing the toggle on the options page, this does not work

Issue #347, opened 2023-10-19
Change profile should invalidate the cache


MrMamen:
### Expected Behavior

When changing profile on a streaming service, trakt-scrobbler should display history for that profile

### Current Behavior

When changing profile on a streaming service, trakt-scrobbler displays cached history for the last profile that was synced.

### Steps to Reproduce

View history on trakt-scrobbler for Netflix
Change to another profile on Netflix.com
View history again on trakt-scrobbler for Netflix
Notice that the history is not correct for this profile

Issue #148, opened 2022-03-17
Viaplay not working


bart268:
I am glad I found this trakt Scrobbler, awesome work.

I have it running with Netflix and HBO Max, only thing I don't get Viaplay to work. When I try to sync Viaplay I get this error:

===
Not Logged In
Looks like you're not logged in to this service. Please go to the page below and log in before trying to sync.

https://viaplay.com/

===

When I click the link, it opens Viaplay where I am logged in so that looks fine. Maybe it has something to do with the region (I'm from The Netherlands)?
I also enabled Rollbar hopefully this gives some information?

MrMamen:
What is the url to Viaplay in your region?

bart268:
The login page: https://viaplay.com/nl-nl/
After login: https://viaplay.com/

MrMamen:
In my region I can see my watch history on this URL:
https://viaplay.no/watched

In the developer tool I can see that it calls this URL to fetch the data:
https://content.viaplay.no/pcdash-no/watched?nextMediaGuid=20455732&nextTimestamp=1589310939561&profileId=E7CA2512C12

Could you provide me with your URL for watch data?

bart268:
I am not fully into the browser development tools but I think this is it.
https://viaplay.com/watched

https://content.viaplay.com/pcdash-nl/watched?profileId=IZmnbwWapGd9m7EKXgqYniIddQabTW17&language=nl

When I open this last URL I get the watched page in something like json. Looking in that data I can see this URL:
https://content.viaplay.com/pcdash-nl/watched?nextMediaGuid=20717488&nextTimestamp=1646942868233

bart268:
I did build the extension with the changes you made. The redirect goes to https://viaplay.com/nl-nl which let me login, however everytime I click the link to sync viaplay history I get logged out again?
If I don't click the link to sync viaplay I can restart the browser without getting logged out.

MrMamen:
Then I haven't been able to fix the issue. On the extension page in chrome (where you install extenstions) could you click on "background" page to open DevTools, and click on the Network-tab. Then try to sync viaplay. See if any of the requests turns red in DevTools. The two I would expect should be something like this:
https://login.viaplay.no/api/persistentLogin/v1?deviceKey=pcdash-no
https://content.viaplay.no/pcdash-no/watched

Which (if any of these fail)? What are your version of these URLS.

Also if opening this URL in your browser:
[...]

PR #369, opened 2024-02-08
Fix Viaplay profile usage

### LINKED ISSUE #155: Viaplay - Wrong profile
When trying to scrobble (manually) from Viaplay it selects the wrong profile (the owner). 

I'm logged into my own profile but I can't change it.

If I hover over the: 

"History for Viaplay, profile: <insert name>"

it says:

"... swich profiles in the service page"

Where exactly is this "service page" ?

lindvall:
Fixes #155

I've implemented support for Viaplay profiles.

Currently it requires clearing caches after changing profile in Viaplay in order to remove the history of the previously selected profile. Any suggestions are welcome, or simply leave it as is.

### DIFF
--- a/src/services/viaplay/ViaplayApi.ts
+++ b/src/services/viaplay/ViaplayApi.ts
@@ -5,6 +5,23 @@ import { Shared } from '@common/Shared';
 import { Utils } from '@common/Utils';
 import { EpisodeItem, MovieItem, ScrobbleItem, ScrobbleItemValues } from '@models/Item';
 
+export interface ViaplayGlobalObject {
+	storeState: {
+		profiles: {
+			activeProfile: string | null;
+			profileList: [
+				{
+					id: string;
+					name: string | null;
+				}
+			];
+		};
+	};
+	userData: {
+		firstName: string | null;
+	};
+}
+
 export interface ViaplayAuthResponse {
 	success: boolean;
 	userData?: {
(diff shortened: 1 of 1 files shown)


src/services/viaplay/ViaplayApi.ts
```diff
@@ -94,11 +119,13 @@ export interface ViaplayProductUserInfo {
 }
 
 class _ViaplayApi extends ServiceApi {
-	INITIAL_URL = 'https://viaplay.com/requirements';
+	INITIAL_URL = 'https://viaplay.com/';
```
MrMamen, line 122: While this may be no longer be correct, it was introduced because of #148
Not sure why you change it, but  perhaps it could be updated to something not broken, like https://viaplay.com/devices ?

MrMamen (approved):
I'm slightly worried that this may break some regions. The profile change is allerede a known bug #347

src/services/viaplay/ViaplayApi.ts
lindvall: The /requirements url was pointing to a 404, which is why I changed it. That may have been a mistake though, as some regions may require a path for the login to work? Let's see if any users experience issues with the change.

lindvall:
Thanks for the approval and the pointer to the catching issue while changing profile! Seems like a rather low priority bug (as you normally only use/sync your own profile).

Hopefully I can solve solve some other issues, and/or implement suggestions in the future.

PR #383, opened 2024-06-02
Added Amazon Australia to allow for Australian Prime Video customers …

nextux:
…to login and sync

Users with Amazon.com.au were unable to sync.  Fixing this by adding in the correct hostPatterns entry for the Australian Amazon site.


MrMamen (approved):

MrMamen:
Thanks!

Issue #395, opened 2024-09-05
Netflix Sign in not working


Nickspongebobpersonal:
I'm Signed in to Netflix, Scrobble, sync and automatic sync is on, but the extension doesn't work and keeps saying I'm not signed in.

r3dicalnusz:
Check #394

r3dicalnusz:
Fixed in v0.4.1

joaociocca:
did it stop working again? I've got 0.4.1 installed and netflix still won't recognize that I'm logged in...
![image](https://github.com/user-attachments/assets/d53b9e39-f382-4f4d-a4ac-47ea8ae4cabf)
![image](https://github.com/user-attachments/assets/8733ef58-efa0-408d-a671-801c787fd31e)
![image](https://github.com/user-attachments/assets/f43be86a-ec8f-4460-bd24-9c494c076777)

r3dicalnusz:
@joaociocca Works fine on my end. Is website access to netflix granted for the extension? Maybe toggle scrobbling and syncing off and on to trigger the query again.

joaociocca:
huh. it seems the permission was removed somewhere since I last used it. weird. Is it possible to implement a check for it on the plugin?

MrMamen:
Closing as this was fixed in v0.4.1.

PR #399, opened 2024-09-17
(fix): broken netflix account regex

### LINKED ISSUE #394: Netflix sync is broken again
### Expected Behavior

Netflix sync to Trakt Watched items

### Current Behavior

Unable to sync with Netflix, but instead showing "Not Logged In" message.  This is somewhat related/similar to the previous issue #384 and #385 

image

It seems Netflix has changed the watch history link again, and basing on the changes, the fix is not just a change in the ACTIVATE_URL variable but also need to track the profile ID (hashed)

TDanks2000:
Fixes the broken "User not logged in" by updating the regex in the netflixapi.ts file. fixes #394 #395


r3dicalnusz (approved):

MrMamen (approved):

MrMamen:
Thanks for fixing this!

Issue #424, opened 2025-07-02
Netflix Sync stopped working


AcidSlide:
### Expected Behavior

Able to list Netflix watch history from the Universal Trakt Scrobbler that needs to be synced

### Current Behavior

Error: "Failed to load history" upon clicking "Netflix" on the "Select a streaming service to sync from:"

### Steps to Reproduce

1. Click the Universal Trakt Scrobbler extesnion
2. Click the "History" icon
3. Click "Netflix" on the available streaming services

![Image](https://github.com/user-attachments/assets/e51dac2b-656b-4144-adc2-604fc683c9f4)

Below is from the browser console:
```
[UTS] Failed to load history. Error: {"request":{"url":"https://www.netflix.com/api/shakti/mre/viewingactivity?languages=en-US&authURL=<SNIPPED-FOR-SECURITY>\x3D\\x3D&pg=0","method":"GET","cancelKey":"default","withHeaders":{},"signal":{}},"status":-1,"isCanceled":false}
history.js:2:392013
```

Note: Tried both RELEASE (v0.4.1) and [BETA (2025-06-24T15:53:17.474Z)](https://github.com/trakt-tools/universal-trakt-scrobbler/releases/tag/beta)

### Additional Information

- OS: Ubuntu 22.04.5 LTS
- Browser: Firefox 140.0.2 (64-bitr)
- Country: US

mojo17:
I can confirm this issue on my end as well. If I try to go to the viewing activity URL  https://www.netflix.com/api/shakti/mre/viewingactivity , I get the response "BLOCKED".

AcidSlide:
It looks like Netflix changed the API for getting the view activities again.  Below is what the new API call looks like:

`https://www.netflix.com/api/aui/pathEvaluator/web/^2.0.0?method=call&callPath=["aui","viewingActivity",1,50]&falcor_server=0.1.0`

The profile id (guid) is in the form data in the payload plus there a lot of custom netflix request headers but not sure if most are new.

GitHubStephan:
Same error on my side

tomfrontman:
same error

rogdex24:
hi,
I made a PR #426 to fix this
please check

cooldude2200:
Any hopes the issue will get updated.

tomfrontman:
no update yet?

DrSeussFreak:
[...]

PR #426, opened 2025-07-07
fix: New Netflix API history fetching logic.

rogdex24:
using the new endpoint for history collection: 

`https://www.netflix.com/api/aui/pathEvaluator/web/^2.0.0?method=call&callPath=["aui","viewingActivity",1,50]&falcor_server=0.1.0`

- Also the reason i have added MAX_RETRIES upto 10 is the netflix api just fails randomly and then succeeds ? so this was the legible solution i found for it

### DIFF
--- a/src/services/netflix/NetflixApi.ts
+++ b/src/services/netflix/NetflixApi.ts
@@ -1,3 +1,4 @@
+/* eslint-disable @typescript-eslint/no-unsafe-assignment */
 import { NetflixService } from '@/netflix/NetflixService';
 import { ServiceApi, ServiceApiSession } from '@apis/ServiceApi';
 import { Requests } from '@common/Requests';
@@ -20,6 +21,7 @@ export interface NetflixGlobalObject {
 				data: {
 					authURL: string;
 					name: string | null;
+					userGuid?: string;
 				};
 			};
 			serverDefs: {
@@ -39,6 +41,7 @@ export interface NetflixPlayerState {
 
 export interface NetflixSession extends ServiceApiSession {
 	authUrl: string;
+	userGuid?: string;
 }
 
 export interface NetflixScrobbleSession {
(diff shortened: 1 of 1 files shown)


rogdex24:
fixes #424

sheharyarn:
Great to see a fix on the way -- Thank you! Hope this is merged soon!

MrMamen:
The call to Netflix seems to work, but when I test this the trakt-api results in 500 errors. Might be a temporary hickup. Need to look more into this later.

GitHubStephan:
any news for the issue? sounds like it is solved now with the fix. Any possibility the use the new fix to try out?

execsumo:
Looks like the Netflix API update works now, but the trakt.tv API call to find matching show/episode fails. With the help of Claude, I built @rogdex24 's fork with the update and tried sync'ing Netflix. The console and network dev tools show that the Netflix API returns my history, but from there the https://api.trakt.tv/search/episode?query= string fails with the trakt-api. Tried the call with Postman and also got 500 error similar to @MrMamen. Think it has something to do with needing to search show first, then episode.

It's beyond my skillset, but thought I'd share.

GitHubStephan:
Still not out? Do you think problem will be solved soon or will it be a bigger thing?
Alternative to sync Netflix with Trakt will be a VIP Paywall in Trakt. Does anyone know other solutions? Its only needed one time for the missed weeks now. Afterwards i can do it manually by App.

rogdex24:
i don't get what's the issue on here , this is working fine and i'm getting it all synced daily 
image

anks-mrangry:
@rogdex24 Scrobbling is working as expected; the issue lies with syncing. For example, we're unable to pull Netflix history to sync watched content from a phone or tablet, consistently receiving a "**Failed to load history**" error.
Screenshot 2025-07-25 195604

jerone:
@rogdex24 Do you have an build XPI that we can try...

MrMamen (approved):

MrMamen:
OK I'm merging this. There might be another problem but this is still an important fix.

eugeniosaintemarie:
Did anyone manage to get the beta _2025-07-29T20:05:16.104Z_ to work?
After a long time like this:
image
The History remains empty

PR #427, opened 2025-07-14
Support Crave UI and API changes

g0tsimid:
Update the service to work with the new Crave UI and GraphQL endpoints.


MrMamen (approved):
I don't have Crave myself, but the code looks good.

Issue #429, opened 2025-07-30
Netflix: Items briefly appear on sync page then disappear after #426


fluteds:
### Expected Behavior

Items should be visible and persist in the sync list until manually dismissed or synced.

### Current Behavior

- Items with information briefly appear on the sync page, then disappear automatically, yet newly watched items are still being synced silently in the background. 
- Previously watched items in history before the update are affected and not synced.
- Options are also greyed out until all history items have flashed and disappeared.

### Steps to Reproduce

- Update to [Beta (2025-07-29T20:05:16.104Z)](https://github.com/trakt-tools/universal-trakt-scrobbler/releases/tag/beta)
- Open the sync page for Netflix
- Items show briefly, then vanish, giving the impression nothing is pending (or has synced)

https://github.com/user-attachments/assets/3256849d-8552-4413-99e6-16cb188b2caf

<img width="1523" height="865" alt="Image" src="https://github.com/user-attachments/assets/6fd64ec2-e2d3-4c77-bdda-887814c00603" />

### Additional Information

Others have mentioned this [issue on the merged PR #426.](https://github.com/trakt-tools/universal-trakt-scrobbler/pull/426#issuecomment-3134611268) The maintainer mentions that there might be an issue as a result of that, but I'm opening an issue to track progress.

- **OS + Version:** macOS Sequoia 15.5 (24F74)
- **Browser + Version:** Firefox Nightly 142.0a1 (2025-07-02) (64-bit)
- **Country/Region:** United Kingdom

eugeniosaintemarie:
The same happend on Chrominium (Vivaldi) for Windows 11 -[screenshot](https://github.com/trakt-tools/universal-trakt-scrobbler/pull/426#issuecomment-3134611268)

DrSeussFreak:
Will add the unchecking "Hide Synced" does not make a difference

keldian:
Anyone try using js breakpoints to stop the items before they disappear? :smile:

jwchia1:
i'm facing the same issue and in the console, i'm getting a whole bunch of these:

`[UTS] Failed to execute action. e: {"request":{"url":"https://api.trakt.tv/shows/215309/seasons/2/episodes/1?extended=full","method":"GET","cancelKey":"default","withHeaders":{"trakt-api-key":"SECRET","trakt-api-version":"2","Authorization":"Bearer SECRET"},"signal":{}},"status":-1,"isCanceled":false}
[...]

PR #431, opened 2025-08-09
Fix for Netflix shows disappearing issue

mlibbey:
#429

ESSENTIAL CHANGE:
- Fixed hiding logic in HistoryList.tsx checkHiddenSelected function
- Changed: const shouldHide = doHide && && isSelectable;
- Prevents non-selectable items from being hidden during processing

ROOT CAUSE:
Netflix shows disappeared because the hiding logic incorrectly marked non-selectable items (items without Trakt metadata) as hidden, causing them to render with 0 height and become invisible in the virtualized list.

SOLUTION:
The fix ensures items remain visible while they are loading or not yet selectable, allowing time for Trakt metadata to be fetched and processed. Only items that are fully processed AND selectable can be hidden.

### DIFF
--- a/src/modules/history/components/HistoryList.tsx
+++ b/src/modules/history/components/HistoryList.tsx
@@ -197,14 +197,18 @@ export const HistoryList = (): JSX.Element => {
 		for (const item of items) {
 			const doHide = item.doHide();
 			const isSelectable = item.isSelectable();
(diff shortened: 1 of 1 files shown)


mlibbey:
Hmm. It looks like the actual issue is around the Netflix data in the API. I think the Min Percentage filter is the core issue. Will submit a new PR.

PR #433, opened 2025-08-10
Fix for Netflix shows disappearing issue

mlibbey:
#429 

It appears that in many cases (all in my case) that the Netflix api does not include bookmark or duration data. As a result, the calculatedProgress was generally less than the "Minimum percentage watched" setting, and thus hidden from view.

If the data doesn't exist, we should assume its fully watched so that it can be shown to the user to decide to import.


smoothlystable:
I can confirm that this fixed the issue with history sync for me, thanks!

MrMamen (approved):
Thanks for your contribution!

eugeniosaintemarie:
Working! ❤️‍🩹

PR #438, opened 2025-10-07
Add support for discoveryplus - scrobble and sync

erased1091:
Added support for discovery+ following some recent changes. I had just about finished a working service extension based on their old platform, when d+ changed everything.

I have reworked and added both scrobble and sync. If you have lots of viewing history it may take a while for the history tab to load. Mine takes about a minute to load up the full history, so not a huge wait. I am new to typescript, so got this working by hook and crook for a UK-based account.

I've also tried to add regional support but is untested, so any help/assistance with the code would be welcome. Some shows may not correctly match to Trakt, either show names not matching, or some episodes split into multiple parts. Not much I can do about that!

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,6 +73,7 @@ If you want to scrobble / sync from Netflix, this is the only Trakt.tv [plugin](
 |       AMC+        |    ✔️    |  ❌  | -                               |
 |       Crave       |    ✔️    |  ✔️  | -                               |
 |    Crunchyroll    |    ❌    |  ✔️  | Can't identify movies as movies |
+|    discovery+     |    ✔️    |  ✔️  | -                               |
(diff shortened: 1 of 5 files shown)


MrMamen (approved):
Thanks for adding this. I do not have an active subscription, and the service have been removed from my region, however I can still access my data. When I tested my profileName was set to "", so the activation failed. This might not be a possible scenario for active subscribers, so it might not be a bug, but maybe some of the other attributes could have been used instead.

erased1091:
Thanks for merging. D+ profile names can't be null, so hopefully not a bug. May be able to fall back to email address if only one profile, I will take a look. Meantime, been working on some other services (ones I use or can use) like bbciplayer and channel4 so expect another PR soon ;-)

PR #458, opened 2026-02-23
Kino.pub support

scytem:
Adding support of kino.pub (russian unofficial streaming service)


scytem:
Since kino.pub stopped issuing API keys for apps, I used the following API keys https://github.com/ValeraGin/kinopub.webos/blob/master/.env#L3-L4

As far as I understood lots of unofficial clients and integration now utilises API keys that've been issued for Kodi (XBMC) long time ago

src/common/BrowserStorage.ts
```diff
@@ -413,6 +430,11 @@ class _BrowserStorage {
 	 * They are only separated by type, to make it easier to understand the downgrade process.
 	 */
 	async downgrade(version: number) {
+		if (version > 11 && this.currentVersion <= 11) {
+			console.log('Downgrading to v11...');
+			await this.doRemove(['kinoPubAuth'] as unknown as (keyof StorageValues)[], true);
```
MrMamen, line 435: There is not really any need to do this. The upgrade/downgrade is mostly for changes in the data-structures, and permanently removed services. Changing branches will get you in trouble, but that's not what this is for. It's for released versions.

.env.example
```diff
@@ -5,3 +5,5 @@ TMDB_API_KEY=
 CHROME_EXTENSION_ID=
 CHROME_EXTENSION_KEY=
 FIREFOX_EXTENSION_ID=
+KINOPUB_CLIENT_ID=
```
MrMamen, line 8: requrering custom secrets for a streaming service is not something that we support so far. It would either require GUI for entering them, or that the published version is build with these keys. @rafaelgomesxyz please advice.

MrMamen (changes requested):
Thanks for your contriubtion. However I'm not sure what I think of this due to the requirements of secrets. @rafaelgomesxyz

PR #462, opened 2026-03-20
kino.pub scorbbling and history support

dm3ch:
Finishes work started in #458 (by @scytem). Adds full scrobbling and history sync support for kino.pub.

  ## What's new vs #458

  - Rewrites history sync to use the JSON API (`/v1/history`) instead of HTML scraping, making it stable against UI changes
  - Implements the OAuth 2.0 device flow for authentication — no credentials stored, works headlessly
  - Adds auto-sync support
  - Fixes episode ID uniqueness so episodes of the same show don't collide in the Trakt search cache
  - Extracts the original title from kino.pub's combined `"Russian Title / Original Title"` format before searching Trakt

  ## Why the common layer had to change

  Three files outside `src/services/kino-pub/` were touched — each for a specific reason:

  **`src/common/BrowserStorage.ts`** — kino.pub uses its own OAuth tokens (access + refresh token, expiry) that must survive browser restarts. The storage schema is versioned
  (`StorageValuesV11`), so adding `kinoPubAuth` required bumping the version and defining the `KinoPubAuthDetails` type here, where all other persisted auth state lives (e.g. Trakt's
  `auth` key). There is no service-local storage path.

  **`src/common/ScriptInjector.ts`** — The injector builds content-script match patterns from each service's `hostPatterns`. The existing regex only handled wildcard-subdomain patterns
  (`*://*.example.com/*`) and converted them to the injector's internal format. kino.pub uses `*://kino.pub/*` — no subdomain wildcard — which the old regex left untransformed. A one-line
  fix adds handling for the exact-host case so the content script is correctly injected on `kino.pub` pages.

  **`webpack.config.ts`** — The device flow needs to show the user their one-time code in a place they can actually read it. Browser notifications require an explicit permission grant and
  are easy to dismiss. Instead, the extension opens its own page (`kino-pub-auth.html`) that shows the verification URL and code together with a copy button. The extension has no runtime
  template engine — every navigable page must be a pre-built webpack entry with its own JS bundle and HTML file generated by `HtmlWebpackPlugin`. Adding `kino-pub-auth` alongside
  `popup`/`history`/`options` is the only supported path. The auth tab is closed automatically once polling confirms the token exchange.


dm3ch:
@MrMamen as you have initially reviewed #458 and this PR finishes the work started there, can I ask you to review this PR as well?

package.json
```diff
@@ -15,7 +15,8 @@
 		"tsc": "tsc --noEmit",
 		"check": "tsc --noEmit && biome check",
 		"fix": "biome check --write",
-		"prepare": "husky install"
+		"prepare": "husky install",
+		"start-chrome-ext": "open -a 'Google Chrome' --args --load-extension=$(pwd)/src/build/chrome --remote-debugging-port=9222 --user-data-dir=/tmp/uts-chrome-profile"
```
MrMamen: This seems to be some kind of debugging, and is macOS-only as far as I can see? Is it required?

src/services/kino-pub/KinoPubApi.ts
```diff
@@ -0,0 +1,358 @@
...
+
+	private hasTokenExpired(auth: KinoPubAuthDetails): boolean {
+		const now = Utils.unix();
+		return auth.created_at + auth.expires_in < now;
+	}
+
+	private async refreshToken(refreshToken: string): Promise<boolean> {
+		try {
+			const responseText = await Requests.send({
+				url: `${this.API_URL}/oauth2/token?grant_type=refresh_token&client_id=${KINOPUB_CLIENT_ID}&client_secret=${KINOPUB_CLIENT_SECRET}&refresh_token=${refreshToken}`,
```
MrMamen: I assume these query parameters are required, ideally they should go in the body.

src/common/ScriptInjector.ts
```diff
@@ -63,7 +63,10 @@ class _ScriptInjector {
 			)
 			.map((service) => ({
 				matches: service.hostPatterns.map((hostPattern) =>
-					hostPattern.replace(/^\*:\/\/\*\./, 'https?:\\/\\/([^/]*\\.)?').replace(/\/\*$/, '')
+					hostPattern
```
MrMamen, line 66: nice. I think this might fix some other bugs as well.

MrMamen (approved):
Looks good. Conciser removing start-chrome-ext, as it doesn't seem relevant to kino.pub.

package.json
dm3ch: Nice catch, it's not required, so I dropped that

src/common/ScriptInjector.ts
dm3ch: Let's hope you're right :)

src/services/kino-pub/KinoPubApi.ts
dm3ch: Moved them to the body, tested still working fine. TY for a catch

dm3ch:
@MrMamen TY for review. Addressed your comments

PR #468, opened 2026-04-26
fix: update Trakt OAuth callback to handle app.trakt.tv redirect

### LINKED ISSUE #467: Login with Trakt.tv broken: OAuth callback redirects to app.trakt.tv
### Expected Behavior

Clicking \"Login with Trakt.tv\" should open the Trakt authorization page, and after approving, complete the OAuth flow and log the user in.

### Current Behavior

After authorizing on Trakt's website, login silently hangs forever and the user is never logged in.

### Root Cause

Trakt changed their server-side behavior: visiting \`https://trakt.tv/apps?code=...\` (the registered OAuth redirect URI) now issues a 301 redirect to \`https://app.trakt.tv/?code=...&mode=media\`.

The extension's content script (\`trakt.js\`) is registered to match \`*://*.trakt.tv/apps*\`. Because Chrome follows server-side redirects before running content scripts, the script never fires on the final \`app.trakt.tv\` URL. The \`finishManualAuth\` callback is never called, so the login Promise never resolves.

A secondary bug was also discovered: \`getCode()\` used \`split('=')[1]\` to extract the authorization code, which breaks when Trakt appends \`&mode=media\` — returning \`"<code>&mode"\` instead of just the code value.

### Steps to Reproduce

1. Install the extension (development or production build)
2. Log out (or clear auth from storage)
3. Click \"Login with Trakt.tv\" in the popup
4. Authorize the app on Trakt's website
5. Login never completes — the extension stays on the login screen

### Additional Information

- **OS + Version:** macOS 25.4.0
- **Browser + Version:** Chrome (Manifest V3)
- **Country/Region:** N/A (reproducible everywhere)

The fix is two changes:
- Add \`*://app.trakt.tv/*\` to the content script \`matches\` in \`webpack.config.ts\`
- Fix \`getCode()\` in \`TraktAuth.ts\` to use \`URLSearchParams\` instead of fragile string splitting

dm3ch:
Fixes #467

## Summary

Trakt changed their server-side OAuth redirect behavior: `https://trakt.tv/apps?code=...` now 301-redirects to `https://app.trakt.tv/?code=...&mode=media`. Because Chrome follows server-side redirects before injecting content scripts, `trakt.js` (matched on `*://*.trakt.tv/apps*`) never fires on the final URL — leaving login permanently pending.

- **`webpack.config.ts`**: Add `*://app.trakt.tv/*` to the content script `matches` so `trakt.js` runs on the new OAuth callback destination
- **`src/apis/TraktAuth.ts`**: Fix `getCode()` to use `URLSearchParams` instead of `split('=')[1]`, which was returning `"<code>&mode"` when Trakt appended `&mode=media` to the redirect URL

## Test plan

- [ ] Log out of the extension (or clear `auth` from `chrome.storage.local`)
- [ ] Click "Login with Trakt.tv" in the popup
- [ ] Authorize the app on Trakt's website
- [ ] Confirm login completes and the popup shows the logged-in state

### DIFF
--- a/src/apis/TraktAuth.ts
+++ b/src/apis/TraktAuth.ts
@@ -47,7 +47,8 @@ class _TraktAuth extends TraktApi {
 	}
 
 	getCode(redirectUrl: string): string {
-		return redirectUrl.split('?')[1].split('=')[1];
+		const search = redirectUrl.includes('?') ? redirectUrl.split('?')[1] : redirectUrl;
+		return new URLSearchParams(search).get('code') ?? '';
 	}
 
 	hasTokenExpired(auth: TraktAuthDetails): boolean {
--- a/webpack.config.ts
+++ b/webpack.config.ts
@@ -246,7 +246,8 @@ const getManifest = (browserName: string, isDev: boolean): string => {
 		content_scripts: [
 			{
 				js: ['trakt.js'],
-				matches: ['*://*.trakt.tv/apps*'],
+				// Trakt redirects oauth callback from trakt.tv/apps to app.trakt.tv
+				matches: ['*://*.trakt.tv/apps*', '*://app.trakt.tv/*'],
 			},
 		],
 		default_locale: 'en',


MrMamen (approved):
While I can't replicate the error, the code seems to be an improvement anyways.

MrMamen:
Thanks for your contribution

PR #457, opened 2026-02-21
Fix: Netflix live scrobble function

sitz:
Fixes #434 
Fixes #442

PR #469, opened 2026-05-29
Throttle Netflix item lookups and cache playback injection

sitz:
Follow-up to #457, addressing the review concerns @MrMamen raised on the merged PR. Rebased onto current master; still relevant after #475 (see below).

## Summary
- **Throttle unresolvable IDs.** The base `ScrobbleParser.parseItemFromApi` blacklists a failing ID via `failingId` so it isn't re-requested. The Netflix override drops that guard, so an ID that won't resolve triggers `getItem` (metadata requests + a history fetch) on every parse tick (~every 500ms). Instead of permanently blacklisting — which would reintroduce the live-content bug, since those IDs resolve only once available — failing IDs are now retried on a 30s interval.
- **Cache the playback injection.** Playback and item-ID are both derived from the same injected `netflix-playback` function, which was injected multiple times per parse cycle. They now share a single injection via a short (250ms) cache that still refreshes on the next tick.

Both changes are self-contained in `NetflixParser`; no base-class behavior is altered for other services.

## Still needed after #475
#475 swapped the metadata endpoint but did not add any throttling to the live-scrobble path:
- `NetflixParser.parseItemFromApi` has no failing-ID guard (it overrides the base parser's `failingId` blacklist), and
- the live lookup `getItem → getItemFromMetadata → fetchSingleMetadata` is uncached (the new `metadataCache` only covers the history path) and now tries up to 4 URL variations per call.

So an unresolvable/live ID still hits ~4 metadata GETs + a history fetch on every parse tick. This PR addresses exactly that flooding without reintroducing the #457 live-content bug.

## Test plan
- [x] `tsc --noEmit` passes
- [x] `biome check` passes
- [x] `build-dev` compiles
- [x] Manual: on-demand show and movie scrobble correctly — play/pause/resume detected and item resolved (show → season/episode/title, movie → title/year) via the `release/metadata` endpoint
- [x] Manual: rebased onto current master; no conflicts (MERGEABLE)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/services/netflix/NetflixParser.ts
+++ b/src/services/netflix/NetflixParser.ts
@@ -1,12 +1,39 @@
 import { NetflixApi, NetflixInjectedPlayback } from '@/netflix/NetflixApi';
 import { ScriptInjector } from '@common/ScriptInjector';
-import { ScrobbleParser } from '@common/ScrobbleParser';
+import { ScrobblePlayback, ScrobbleParser } from '@common/ScrobbleParser';
+import { ScrobbleItem } from '@models/Item';
 
 class _NetflixParser extends ScrobbleParser {
+	/**
+	 * How long (in ms) to wait before retrying `getItem` for an ID that failed to resolve.
+	 *
+	 * Netflix live content isn't always available in the metadata API right away, so unlike
+	 * the base parser - which blacklists a failing ID for the rest of the session - we keep
+	 * retrying so the item can resolve once it becomes available. The retry is throttled to
+	 * avoid hammering the API on every parse tick (which runs roughly every 500ms).
+	 */
+	private static readonly FAILING_ID_RETRY_INTERVAL = 30 * 1e3;
+
+	/**
+	 * How long (in ms) a single playback injection is reused for. Both the playback and the
+	 * item ID are derived from the same injected function, so caching it for a short window
+	 * collapses the multiple injections that happen within a single parse cycle into one,
+	 * while still refreshing on the next tick.
+	 */
+	private static readonly PLAYBACK_CACHE_TTL = 250;
+
+	private failingIds = new Map<string, number>();
+	private cachedPlayback: NetflixInjectedPlayback | null = null;
+	private cachedPlaybackAt = 0;
+
 	constructor() {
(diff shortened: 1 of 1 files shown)


MrMamen (approved):

MrMamen:
Seems OK to me, but I'll await merge as long as it is marked draft.

MrMamen:
This might not be relevant after #475. Please rebase/merge from master and see if this is still needed.

sitz:
Rebased onto current master.

This is still needed after #475 — that PR swapped the metadata endpoint but didn't add throttling to the live-scrobble path:

- `NetflixParser.parseItemFromApi` has no failing-ID guard (it overrides the base parser's `failingId` blacklist), and
- the live lookup `getItem → getItemFromMetadata → fetchSingleMetadata` is uncached (the new `metadataCache` only covers the history path) and now tries up to 4 URL variations per call.

So an unresolvable/live ID still triggers ~4 metadata GETs + a history fetch on every parse tick (~500ms). This PR throttles those failing-ID retries to a 30s interval (instead of permanently blacklisting, which would reintroduce the live-content bug from #457) and collapses the repeated playback injections within a parse cycle into a single 250ms-cached injection. Both changes are self-contained in `NetflixParser`.

Verified: base scrobble behavior works end-to-end for a show and a movie (play/pause/resume + item resolution via the `release/metadata` endpoint), `tsc --noEmit` and `biome check` pass, `build-dev` compiles.

anks-mrangry:
Hi @MrMamen,

There have been a lot of great fixes pushed to the tool over the past few months, but they currently seem to be available only in the beta branch. Do you have a rough ETA on when these updates will be packaged into an official store release? The last public release was v0.6.0 back on November 19, 2025.

Thanks for all your hard work on this extension!

MrMamen:
@rafaelgomesxyz is the only maintainer with access to the release. I'll see if I can make him release shortly.

rafaelgomesxyz:
The release script has been fixed. The new version is already available in the Firefox store, but the Chrome store has a review process, so it takes longer to be available there.

PR #459, opened 2026-02-26
Improve episode matching and error handling for Crunchyroll

damman01:
## Summary

This PR enhances episode matching capabilities and improves error handling, particularly for anime content and services using absolute episode numbering.

## Issues Fixed

Fixes #441 - Movies from crunchyroll are not trackable
Fixes #330 - Crunchyroll history sync fails due to title mismatch
Fixes #321 - When episode number does not match, episode name could be used as fallback
Fixes #315 - Crunchyroll sync cannot load history

### Core Model Enhancement
- **Add isAbsolute field to EpisodeItem** - Track episodes using absolute numbering (common in anime)

### Error Handling Improvements
- **Add Cloudflare rate limiting detection** - Automatically detect and retry when encountering Cloudflare Error 1015 and other rate limit responses

### TMDB Integration
- **Add TV show search method** - Search for shows with optional year filtering and anime prioritization for Crunchyroll
- **Add episode title search** - Find episodes within a show by matching episode titles

### Episode Search Enhancements
- **Absolute episode number resolution** - Convert absolute episode numbers to season/episode pairs using Trakt season data
- **TMDB fallback mechanism** - When Trakt episode lookup fails, search TMDB for the show and find the episode by title
- **Edge case handling** - Detect new seasons not yet in Trakt database

### Crunchyroll Service Improvements
- **Movie detection** - Distinguish between movies and episodes in Crunchyroll history
- **Enhanced metadata parsing** - Use sequence_number as fallback, strip language suffixes (Dub/Sub)
- **Absolute numbering support** - Mark episodes with isAbsolute flag for proper handling
- **Add updateItemFromHistory() stub** - Resolve missing method implementation error

## Benefits

- ✅ Movies from Crunchyroll are now properly detected and trackable
- ✅ Episode title matching as fallback when season/episode numbers don't align
- ✅ Handles translation mismatches between Crunchyroll and Trakt/TMDB
- ✅ Better matching accuracy for anime content using absolute episode numbering
- ✅ More resilient to API rate limiting and temporary failures
- ✅ Handles edge cases with new seasons and absolute episode numbering
- ✅ Improved movie vs episode detection for streaming services

## Testing

These changes have been tested with Crunchyroll history parsing and various anime titles using absolute episode numbering.

---

*Co-authored-by: damman01 <bigdome91@gmail.com>*

Issue #460, opened 2026-02-27
Unable to Match Episodes (Crunchyroll)


schellbell:
### Expected Behavior

Crunchyroll titles have always had some difficulty matching (especially when their TMDB titles were just "Episode X", but usually it just required one or two corrections

### Current Behavior

Recently TraktUniversalScrobbler has been unable to match any series with any episode title at all. I have tried logging out and logging back in, clearing all caches, incognito mode, and removing/reinstalling the extension.

<img width="276" height="888" alt="Image" src="https://github.com/user-attachments/assets/ccd5e0c8-2d7c-4ae0-b84a-f82d81fd066b" />

### Additional Information

- **Windows 11**
- **Chrome 145.0.7632.117**
- **Canada**

damman01:
@schellbell try the curent beta release, it should find the episodes from crunchyroll too
[Beta (2026-03-04T11:41:04.636Z)](https://github.com/trakt-tools/universal-trakt-scrobbler/releases/tag/beta)

schellbell:
Alright just tried it -- did a full removal of Trakt/all Trakt-related files (or at least as many as I could find), cleared caches, then installed the beta. Unfortunately, same issue. Though I'm not sure if I screwed something up or not as it is showing the version as 0.6.0

nomd346:
same issue

im going to try the beta hopefully it works

nomd346:
yea the beta worked perfectly for me and yes it shows the version as 0.6.0

nomd346:
ok i will say that it sometimes doesnt detect that i watched an episode which is weird

this has happened to me 2 times now

excludehk:
how to download the beta? it only says source code zip

PR #472, opened 2026-06-09
Gate expensive episode fallbacks behind absolute numbering flag

MrMamen:
The TMDB and absolute-numbering fallbacks added in #459 ran for every failed episode lookup from every service, costing several extra Trakt requests per item. During history sync with many unmatched items this can trigger Trakt's rate limit.

Introduce EpisodeItem.isAbsolute (described in the #459 commit message but never actually added) and only run the fallbacks when it's set. Crunchyroll, which numbers anime episodes sequentially across seasons, marks its episodes with the flag, preserving the behavior #459 was added for.


MrMamen:
The fallbacks remain active for Crunchyroll, which they were added for #460

PR #475, opened 2026-06-14
Replace removed Netflix pathEvaluator metadata endpoint

### LINKED ISSUE #474: Netflix sync not working at all
Opening the history only gives a red "Failed to load history" banner in the bottom left corner. I tried clearing caches and data, signing off and on to Netflix and Trakt.

Bioll:
## Problem

Closes #474. Netflix removed the bulk `pathEvaluator` endpoint that `NetflixApi.getHistoryMetadata()` relied on. All known route variations now fail:

| Endpoint | Status |
| --- | --- |
| `POST /nq/website/memberapi/<BUILD_IDENTIFIER>/pathEvaluator` | 502 |
| `POST /nq/website/memberapi/release/pathEvaluator` | 412 |
| `POST /api/shakti/<BUILD_IDENTIFIER>/pathEvaluator` | 421 |
| `POST /api/shakti/mre/pathEvaluator` | 404 |

The history list itself still loads via the AUI endpoint (#426), but because `getHistoryMetadata()` threw, the whole load failed with the red "Failed to load history" banner even though the items had already been fetched.

## Fix — metadata endpoint

The single metadata endpoint still works through the `release` route:

```
GET https://www.netflix.com/nq/website/memberapi/release/metadata?languages=en-US&movieid=<id>
```

For a show id it returns the whole structure (all seasons/episodes with `id`/`seq`/`title`, plus `year` and `hiddenEpisodeNumbers`), so one request per unique show/movie on a history page rebuilds everything `getHistoryMetadata()` needs.

- `getHistoryMetadata()` no longer POSTs to `pathEvaluator`; it collects the unique show/movie ids and fetches each via the single metadata endpoint, caching responses and remembering the working URL variation so later pages don't re-probe.
- Episode history items are located by id inside the returned show structure to recover `season.seq` / `episode.seq` / `releaseYear` / `hiddenEpisodeNumbers`.
- If metadata can't be fetched at all, items are returned without it (matched on Trakt by title) instead of throwing, so the history still loads.
- The same `fetchSingleMetadata()` path also backs the live-scrobble `getItem` flow.

The GraphQL route (`web.prod.cloud.netflix.com/graphql`) was considered but relies on persisted-query IDs that bump on every web-client update and is keyed by `seasonId` (not present in history data), so the single `release/metadata` route is simpler and reuses the existing `parseMetadata` structure.

## Fix — escaped `authURL`

Values captured by the `extractSession` regex fallback from the `/settings/viewed/` HTML keep their JS string escapes, so the trailing `==` of `authURL` arrives as `\x3D\x3D` and corrupts any request that includes it. This silently broke the `&authURL=…` metadata variation during background/auto syncs with no Netflix tab open. Added `decodeJsString()` to decode `\xNN`, `\uNNNN`, and `\/` for `authURL` and `profileName`. The injected-session path is unaffected (it reads decoded values straight off the `netflix` object).

## Testing

Built from current `master` + this patch: Netflix history loads again with correct years and season/episode numbers, and the banner is gone. `tsc --noEmit` and `biome check` pass.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/services/netflix/NetflixApi.ts
+++ b/src/services/netflix/NetflixApi.ts
@@ -41,6 +41,7 @@ export interface NetflixPlayerState {
 export interface NetflixSession extends ServiceApiSession {
 	authUrl: string;
 	userGuid?: string;
+	buildIdentifier?: string;
 }
 
 export interface NetflixScrobbleSession {
@@ -98,6 +99,7 @@ export type NetflixMetadataItem =
 
 export interface NetflixMetadataEpisodeItem {
 	releaseYear: number;
+	title?: string;
 	summary: {
 		episode: number;
 		id: number;
@@ -116,6 +118,7 @@ export type NetflixMetadataEpisodeItemWithSeason = NetflixMetadataEpisodeItem & 
 };
 
 export interface NetflixMetadataShowItem {
+	title?: string;
 	seasonList: {
 		current: ['seasons', string];
 	};
@@ -123,6 +126,7 @@ export interface NetflixMetadataShowItem {
 
 export interface NetflixMetadataMovieItem {
 	releaseYear: number;
+	title?: string;
 	summary: {
 		id: number;
 	};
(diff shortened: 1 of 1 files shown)


MrMamen (approved):
Thanks!

PR #483, opened 2026-07-11
Add Animationdeliverynetwork as service

Superritchman:
Implementation for the french anime streaming service ADN.

ADN is avaible in france, germany and poland.
https://animationdigitalnetwork.com/
https://en.wikipedia.org/wiki/M%C3%A9dia-Participations#Animation_Digital_Network

### DIFF
--- a/src/services/adn/AdnApi.ts
+++ b/src/services/adn/AdnApi.ts
@@ -200,7 +200,9 @@ class _AdnApi extends ServiceApi {
 
 	updateItemFromHistory(item: ScrobbleItemValues, historyItem: AdnHistoryItem): void {
 		item.watchedAt = Utils.unix(historyItem.user.watchDate);
-		item.progress = 100;
+		item.progress = historyItem.user.isFullyWatched
+			? 100
+			: (historyItem.duration / historyItem.user.stoptime) * 100;
 	}
 
 	convertHistoryItems(historyItems: AdnHistoryItem[]) {
@@ -215,7 +217,9 @@ class _AdnApi extends ServiceApi {
 					imageUrl: historyItem.image,
 					year: Number.parseInt(historyItem.show.firstReleaseYear),
 					watchedAt: Utils.unix(historyItem.user.watchDate),
-					progress: 100,
+					progress: historyItem.user.isFullyWatched
+						? 100
+						: (historyItem.duration / historyItem.user.stoptime) * 100,
 				});
 				items.push(item);
 			} else {
@@ -228,7 +232,9 @@ class _AdnApi extends ServiceApi {
 					imageUrl: historyItem.image,
 					year: Number.parseInt(historyItem.show.firstReleaseYear),
 					watchedAt: Utils.unix(historyItem.user.watchDate),
-					progress: 100,
+					progress: historyItem.user.isFullyWatched
+						? 100
+						: (historyItem.duration / historyItem.user.stoptime) * 100,
 					show: {
 						id: historyItem.show.id.toString(),
 						serviceId: this.id,
--- a/src/services/adn/AdnApi.ts
+++ b/src/services/adn/AdnApi.ts
@@ -1,0 +1,275 @@
+import { ServiceApi, ServiceApiSession } from '@apis/ServiceApi';
+import { Requests } from '@common/Requests';
+import { EpisodeItem, MovieItem, ScrobbleItem, ScrobbleItemValues } from '@models/Item';
+import { AdnService } from '@/adn/AdnService';
+import { ScriptInjector } from '@common/ScriptInjector';
+import { Utils } from '@common/Utils';
+import { Shared } from '@common/Shared';
+
+import browser from 'webextension-polyfill';
+
+export interface AdnSession extends ServiceApiSession {
+	auth: {
+		accessToken: string;
+		refreshToken: string;
+	};
+	profileId: number | null;
+}
+
+export interface AdnTokenData {
+	accessToken: string;
+	refreshToken: string;
+}
+
+export interface AdnHistoryPage {
+	videos: AdnHistoryItem[];
+}
+
+export interface AdnHistoryItem {
+	id: number;
+	duration: number;
+	number: string;
+	shortNumber: string;
+	season: string;
+	releaseDate: string;
+	type: string;
+	image: string;
+	show: {
+		id: number;
+		title: string;
+		originalTitle: string;
+		firstReleaseYear: string;
+	};
+	name: string;
+	title: string;
+	user: {
+		id: number;
+		isFullyWatched: boolean;
+		stoptime: number;
+		watchDate: string;
+	};
+}
+
+class _AdnApi extends ServiceApi {
+	HOST_URL: string;
+	API_URL: string;
+	REFRESH_URL: string;
+	PROFILE_URL: string;
+	HISTORY_URL: string;
+
+	isActivated: boolean;
+	session: AdnSession | null = null;
+
+	pageSize: number = 25;
+	lang: string;
+
+	request = Requests;
+
+	constructor() {
+		super(AdnService.id);
+
+		this.HOST_URL = `${AdnService.homePage}/de/account/history`;
+		this.API_URL = 'https://gw.api.animationdigitalnetwork.com';
+		this.REFRESH_URL = `${this.API_URL}/authentication/refresh`;
+		this.PROFILE_URL = `${this.API_URL}/user/public/profile`;
+		this.HISTORY_URL = `${this.API_URL}/viewing/history?limit=${this.pageSize}`;
+
+		this.lang = browser.i18n.getUILanguage();
+
+		this.isActivated = false;
+	}
+
+	async activate() {
+		try {
+			const partialSession = await this.getSession();
+			if (
+				!partialSession ||
+				!partialSession.auth ||
+				!partialSession.auth.accessToken ||
+				!partialSession.auth.refreshToken
+			) {
+				throw new Error();
+			}
+
+			const verifyTokenResponse = await Requests.send({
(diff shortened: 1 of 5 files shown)


Superritchman:
I added a fixed list of available distributions (fr, de, pl). getUILanguage determines which distribution header to send, so responses match the user's language.
If the UI language doesn't match any distribution, "fr" is set as fallback (ADN's native language). Since english isn't available.

src/services/adn/AdnApi.ts
```diff
@@ -0,0 +1,321 @@
...
+				method: 'GET',
+				headers: {
+					Authorization: `Bearer ${partialSession.auth.accessToken}`,
+					'x-profile-id': `${partialSession.profileId}`,
+					'x-source': 'Web',
+					'x-target-distribution': `${this.lang}`,
+				},
+			});
+
+			if (verifyTokenResponse === '{"message":"Unauthorized"}') {
```
MrMamen: `Requests.send()` throws a `RequestError` for any status >= 400 (see `Requests.ts`), so a 401 from the profile endpoint will never reach this string comparison — it will land in the outer `catch` instead, and the token refresh never runs. In practice an expired access token means activation silently fails. You'll want to catch the `RequestError`, check `err.status === 401`, and run the refresh flow from there. Comparing against an exact JSON string would also be brittle even if the API returned it with a 200.

src/services/adn/AdnApi.ts
```diff
@@ -0,0 +1,321 @@
...
+					profileId: partialSession.profileId ?? null,
+				};
+			}
+
+			this.isActivated = true;
+		} catch (_err) {}
+	}
+
+	async checkLogin() {
+		if (!this.isActivated || !!this.session) {
```
MrMamen: This condition is true both when there is no session *and* when there is one (`!!this.session`), so `activate()` runs on every `checkLogin()` call, including a fresh profile request each time. I assume this was meant to be `!this.session`. The other services just use `if (!this.isActivated)`.

src/services/adn/AdnApi.ts
```diff
@@ -0,0 +1,321 @@
...
+						accessToken: partialSession.auth.accessToken,
+						refreshToken: partialSession.auth.refreshToken,
+					},
+					profileName: partialSession.profileName ?? null,
+					profileId: partialSession.profileId ?? null,
+				};
+			}
+
+			this.isActivated = true;
+		} catch (_err) {}
```
MrMamen: Swallowing all errors here hides activation failures from the user and from the logs. The convention in the other services (e.g. `CraveApi`) is:

```ts
} catch (err) {
	if (Shared.errors.validate(err)) {
		Shared.errors.log(`Failed to activate ${this.id} API`, err);
	}
	throw new Error('Failed to activate API');
}
```

src/services/adn/AdnApi.ts
```diff
@@ -0,0 +1,321 @@
...
+
+		this.lang = this.getDistributionLang();
+
+		this.isActivated = false;
+	}
+
+	getDistributionLang() {
+		const uiLang = browser.i18n.getUILanguage();
+		// fallback to france, since this is the native service language (no english available)
+		return SUPPORTED_DIST_LANGUAGES.includes(uiLang) ? uiLang : 'fr';
```
MrMamen, line 94: Building on the hardcoded distribution list you added: `browser.i18n.getUILanguage()` usually returns a full locale like `de-DE` or `fr-FR`, which never matches `['fr', 'pl', 'de']` — so e.g. a German user would still always get the `fr` fallback. Normalizing first, e.g. `uiLang.split('-')[0]`, should make this work as intended.

src/services/adn/AdnApi.ts
```diff
@@ -0,0 +1,321 @@
...
+	async activate() {
+		try {
+			const partialSession = await this.getSession();
+			if (
+				!partialSession ||
+				!partialSession.auth ||
+				!partialSession.auth.accessToken ||
+				!partialSession.auth.refreshToken
+			) {
+				throw new Error();
```
MrMamen: Please give these errors descriptive messages (e.g. `'Invalid session'`), like the other services do — an empty `Error` makes the logs hard to act on. Same for the one in `loadHistoryItems()` (line 174).

src/services/adn/AdnApi.ts
```diff
@@ -0,0 +1,321 @@
...
+	HISTORY_URL: string;
+	VIDEO_ITEM_URL: string;
+
+	isActivated: boolean;
+	session: AdnSession | null = null;
+
+	pageSize: number = 25;
+	lang: string;
+
+	request = Requests;
```
MrMamen, line 74: This field is only used in some places while other calls use `Requests` directly, so the file is inconsistent with itself. None of the other services alias it — I'd drop the field and call `Requests.send()` everywhere.

src/services/adn/AdnApi.ts
```diff
@@ -0,0 +1,321 @@
...
+					url: `${this.REFRESH_URL}`,
+					method: 'POST',
+					headers: {
+						'Content-Type': 'application/json',
+						'x-access-token': partialSession.auth.accessToken,
+						'x-profile-id': `${partialSession.profileId}`,
+						'x-source': 'Web',
+						'x-target-distribution': `${this.lang}`,
+					},
+					body: `{"refreshToken":"${partialSession.auth.refreshToken}"}`,
```
MrMamen, line 138: `Requests` already `JSON.stringify`s non-string bodies, so this can just be `body: { refreshToken: partialSession.auth.refreshToken }` — safer than building JSON by string interpolation.

src/services/adn/AdnApi.ts
```diff
@@ -0,0 +1,321 @@
...
+			headers: {
+				Authorization: `Bearer ${this.session.auth.accessToken}`,
+				'x-target-distribution': `${this.lang}`,
+				'x-profile-id': `${this.session.profileId}`,
+			},
+			method: 'GET',
+			cancelKey,
+		});
+
+		const responseJson = JSON.parse(responseText);
```
MrMamen, line 198: The `AdnHistoryPage` interface is defined but never used — this parse can be typed with it: `JSON.parse(responseText) as AdnHistoryPage`. Right now `responseJson` is `any`.

src/services/adn/AdnApi.ts
```diff
@@ -0,0 +1,321 @@
...
+	getHistoryItemId(historyItem: AdnHistoryItem): string {
+		return historyItem.id.toString();
+	}
+
+	updateItemFromHistory(item: ScrobbleItemValues, historyItem: AdnHistoryItem): void {
+		item.watchedAt = Utils.unix(historyItem.user?.watchDate);
+		item.progress =
+			historyItem.user?.isFullyWatched || !historyItem.user
+				? 100
+				: (historyItem.user.stoptime / historyItem.duration) * 100;
```
MrMamen, line 225: This progress calculation is duplicated in `convertHistoryItems()`. Consider extracting a small helper (see `NetflixProgress.ts` from #487 for prior art). It should also guard against `duration` being 0 (NaN) and clamp the result to 0–100, since `stoptime` can exceed `duration`.

src/services/adn/AdnApi.ts
```diff
@@ -0,0 +1,321 @@
...
+
+	pageSize: number = 25;
+	lang: string;
+
+	request = Requests;
+
+	constructor() {
+		super(AdnService.id);
+
+		this.HOST_URL = `${AdnService.homePage}/de/account/history`;
```
MrMamen, line 79: The injection URL hardcodes the `/de/` path even though the distribution language is otherwise dynamic (with an `fr` fallback). Using `this.lang` here, or just the home page, would be more consistent.

src/services/adn/AdnApi.ts
```diff
@@ -0,0 +1,321 @@
...
+		if (!this.session) {
+			throw new Error('Invalid session');
+		}
+		try {
+			const responseText = await this.request.send({
+				url: `${this.VIDEO_ITEM_URL}/${id}`,
+				method: 'GET',
+			});
+			const response = JSON.parse(responseText) as AdnVideoItemResponse;
+			console.log(response.video);
```
MrMamen: Two things here: the leftover `console.log(response.video)` should be removed, and this request sends no `Authorization`/`x-target-distribution` headers unlike every other call in this file — if the endpoint requires them, the scrobbler's item lookup will fail.

MrMamen (changes requested):
Thanks for the contribution! The overall structure follows the existing service template nicely (Service/Api/Parser/entry file, README entry, minimal parser), and lint/typecheck pass. However, there are a few functional bugs in the activation/login flow and some deviations from the conventions used by the other services, so I'm requesting changes. See the inline comments for details. Please note that some of the minor issues are not required, I just kept them as this review was done in collaboration with Claude AI.

src/services/adn/AdnApi.ts
```diff
@@ -0,0 +1,326 @@
...
+	async loadHistoryItems(cancelKey = 'default'): Promise<AdnHistoryItem[]> {
+		let historyItems: AdnHistoryItem[] = [];
+
+		if (
+			!this.session ||
+			!this.session.auth ||
+			!this.session.auth.accessToken ||
+			!this.session.auth.refreshToken
+		) {
+			throw new Error();
```
MrMamen: This Error still has no message, 'Invalid session' would match the other services

src/services/adn/AdnApi.ts
MrMamen: console.log is still here. I think it would not be needed in production?

MrMamen (approved):
Thanks for all the fixes. Two trivial leftovers below I would prefer to have fixed before merging.

PR #485, opened 2026-07-18
Add kino.watch domain for Kinopub

dm3ch:
## Summary
- update Kinopub homepage to https://kino.watch/
- add kino.watch host patterns so the content script runs on the new domain
- keep kino.pub host patterns for backward compatibility

## API notes
- keep the existing Kinopub API host (`https://api.service-kp.com`) because the OAuth/device and history endpoints still respond there
- the device-flow endpoint returns `https://kino.watch/device` as the verification URL

## Checks
- pnpm tsc
- pnpm exec biome check src/services/kino-pub/KinoPubService.ts src/services/kino-pub/KinoPubApi.ts

Note: full `pnpm check` currently fails on ignored local file `.claude/settings.local.json`, unrelated to this PR.

### DIFF
--- a/src/common/BrowserStorage.ts
+++ b/src/common/BrowserStorage.ts
@@ -14,7 +14,7 @@ import '@services';
 import { PartialDeep } from 'type-fest';
 import browser, { Manifest as WebExtManifest } from 'webextension-polyfill';
 
-export type StorageValues = StorageValuesV12;
+export type StorageValues = StorageValuesV13;
 export type StorageValuesOptions = StorageValuesOptionsV4;
 export type StorageValuesSyncOptions = StorageValuesSyncOptionsV3;
 
@@ -24,6 +24,10 @@ export type KinoPubAuthDetails = {
 	expires_in: number;
 	refresh_token: string;
 	created_at: number;
+};
+
+export type StorageValuesV13 = Omit<StorageValuesV12, 'version'> & {
+	version?: 13;
 };
 
 export type StorageValuesV12 = Omit<StorageValuesV11, 'version'> & {
@@ -244,7 +248,7 @@ export type BrowserStorageSetValues = Omit<StorageValues, 'options' | 'syncOptio
 export type BrowserStorageRemoveKey = Exclude<keyof StorageValues, 'options' | 'syncOptions'>;
 
 class _BrowserStorage {
-	readonly currentVersion = 12;
+	readonly currentVersion = 13;
 
 	isSyncAvailable: boolean;
 	options = {} as StorageValuesOptions;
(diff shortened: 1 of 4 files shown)


src/services/kino-pub/KinoPubService.ts
```diff
@@ -2,9 +2,15 @@ import { Service } from '@models/Service';
 
 export const KinoPubService = new Service({
 	id: 'kino-pub',
-	name: 'Kino.pub',
-	homePage: 'https://kino.pub/',
-	hostPatterns: ['*://kino.pub/*', '*://*.kino.pub/*', '*://api.service-kp.com/*'],
+	name: 'Kinopub',
+	homePage: 'https://kino.watch/',
+	hostPatterns: [
+		'*://kino.watch/*',
```
MrMamen, line 8: To confirm and expand on the Copilot comment with the pattern this repo already uses: `saveOptions()` only requests `service.hostPatterns` as origins when the user toggles the service on/off, so existing users who already have Kinopub enabled will never be granted the new `kino.watch` origins after the update — and the content script won't run on the new domain for them.

There's precedent for exactly this situation in `BrowserStorage.ts`: the v12 upgrade step disables kino-pub when `*://api.service-kp.com/*` isn't granted, so that re-enabling the service in the options triggers the permission request (which requires a user gesture). I'd suggest bumping `currentVersion` to 13 and adding a v13 step that does the same when `browser.permissions.contains({ origins: ['*://kino.watch/*', '*://*.kino.watch/*'] })` returns false. (No corresponding downgrade step is needed — v12 doesn't have one either.)

MrMamen (changes requested):
Thanks for the fix! The change itself looks good: keeping the old `kino.pub` patterns and the existing API host makes sense, the parser's URL regex is domain-agnostic so scrobbling will work on the new domain, and the verification URL comes from the API response so nothing else needs updating. However, there's one functional gap around existing users never being granted the new `kino.watch` origins — see the inline comment, which expands on the Copilot comment with the migration pattern this repo already uses. Requesting changes for that.

One more thing while you're at it: Kinopub isn't listed in the README's streaming service table at all (a pre-existing gap, not introduced by this PR) — since you're already touching the naming and domain, it would be great to add a row for it there.

dm3ch:
Addressed the review findings: added the v13 migration that disables enabled Kinopub for users missing the new kino.watch origins, so re-enabling requests them with a user gesture; also added Kinopub to the README table.

MrMamen (approved):

MrMamen (approved):

src/common/BrowserStorage.ts
```diff
@@ -26,6 +26,10 @@ export type KinoPubAuthDetails = {
 	created_at: number;
 };
 
+export type StorageValuesV13 = Omit<StorageValuesV12, 'version'> & {
+	version?: 13;
```
MrMamen, line 30: this might give a conflict with your other branch: #486

PR #486, opened 2026-07-18
Add AniLibria scrobble and sync support

dm3ch:
## Summary
- add support for AniLibria, a Russian anime streaming service, on anilibria.top
- add both playback scrobbling and account history sync for AniLibria episodes
- sync watched AniLibria account timecodes through the existing browser session
- preserve first-import timestamps as approximate watched dates when AniLibria does not provide real watched dates

## Notes
- keeps the approximate watched-date cache service-local in extension local storage
- limits code changes to AniLibria service files plus the README service table

## Validation
- pnpm check
- pnpm build-dev
- fast manual AniLibria smoke test

### DIFF
--- a/src/services/anilibria/AnilibriaApi.ts
+++ b/src/services/anilibria/AnilibriaApi.ts
@@ -1,0 +1,236 @@
+import { ServiceApi } from '@apis/ServiceApi';
+import { Requests } from '@common/Requests';
+import { Shared } from '@common/Shared';
+import { Utils } from '@common/Utils';
+import { EpisodeItem, ScrobbleItem, ScrobbleItemValues } from '@models/Item';
+import { AnilibriaService } from '@/anilibria/AnilibriaService';
+import browser from 'webextension-polyfill';
+
+interface AnilibriaProfileResponse {
+	nickname?: string;
+	login?: string;
+	email?: string;
+}
+
+interface AnilibriaName {
+	main?: string | null;
+	english?: string | null;
+	alternative?: string | null;
+}
+
+interface AnilibriaRelease {
+	id: number;
+	year?: number | null;
+	name?: AnilibriaName | null;
+	alias?: string | null;
+	poster?: {
+		thumbnail?: string | null;
+		preview?: string | null;
+		optimized?: {
+			thumbnail?: string | null;
+			preview?: string | null;
+		} | null;
+	} | null;
+}
+
+export interface AnilibriaEpisodeResponse {
+	id: string;
+	name?: string | null;
+	name_english?: string | null;
+	ordinal?: number | null;
+	duration?: number | null;
+	updated_at?: string | null;
+	release_id: number;
+	release?: AnilibriaRelease | null;
+}
+
+interface AnilibriaRawTimecode extends Array<string | number | boolean | null> {
+	0: string;
+	1: number;
+	2: boolean;
+}
+
+export interface AnilibriaHistoryItem {
+	releaseEpisodeId: string;
+	time: number;
+	isWatched: boolean;
+	firstImportedAt: number;
+}
+
+const HOST_URL = 'https://anilibria.top';
+
+const getBestTitle = (release?: AnilibriaRelease | null): string =>
+	release?.name?.english || release?.name?.main || release?.name?.alternative || '';
+
+const getImageUrl = (release?: AnilibriaRelease | null): string | undefined => {
+	const path =
+		release?.poster?.optimized?.thumbnail ||
+		release?.poster?.thumbnail ||
+		release?.poster?.optimized?.preview ||
+		release?.poster?.preview;
+	if (!path) {
+		return undefined;
+	}
+	return path.startsWith('http') ? path : `${HOST_URL}${path}`;
+};
+
+class _AnilibriaApi extends ServiceApi {
+	HOST_URL = HOST_URL;
+	private hasLoadedHistory = false;
+
+	constructor() {
+		super(AnilibriaService.id);
+	}
+
+	override reset(): void {
+		super.reset();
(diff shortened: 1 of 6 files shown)


src/services/anilibria/AnilibriaApi.ts
```diff
@@ -0,0 +1,236 @@
...
+		const responseText = await Requests.send({
+			url: `${this.HOST_URL}/api/v1/anime/releases/episodes/${releaseEpisodeId}`,
+			method: 'GET',
+			cancelKey,
+		});
+		return JSON.parse(responseText) as AnilibriaEpisodeResponse;
+	}
+
+	private async sendSessionRequest(url: string): Promise<string> {
+		const response = await fetch(url, {
```
MrMamen: Is there a specific reason for using a raw `fetch` here instead of `Requests.send()` like the other services? URLs covered by granted host permissions are treated as same-origin in extension contexts, so session cookies are sent through `Requests.send()` too — that's what e.g. the Netflix sync relies on. The raw `fetch` bypasses the shared request layer, losing `cancelKey`/abort support (sync cancellation won't cancel these requests), the rate-limit/Cloudflare retry handling, and typed `RequestError`s. If there is a reason it doesn't work, it deserves a comment here.

src/services/anilibria/AnilibriaApi.ts
```diff
@@ -0,0 +1,236 @@
...
+		if (!loggedIn) {
+			this.hasReachedHistoryEnd = true;
+			return [];
+		}
+
+		const responseText = await this.sendSessionRequest(
+			`${this.HOST_URL}/api/v1/accounts/users/me/views/timecodes`
+		);
+		const rawTimecodes = JSON.parse(responseText) as AnilibriaRawTimecode[];
+		const { anilibriaFirstImportedAt = {} } = (await browser.storage.local.get(
```
MrMamen: This accesses `browser.storage.local` directly with a key that isn't registered in the storage schema. The convention is to register service-local keys in `StorageValues` (see `kinoPubAuth` in `BrowserStorage.ts`) so they're typed and visible to the storage lifecycle (clearing, up/downgrades). Also worth considering: this cache only ever grows — entries for episodes that disappear from the timecodes response are never pruned.

src/services/anilibria/AnilibriaApi.ts
```diff
@@ -0,0 +1,236 @@
...
+
+	async loadHistoryItems(_cancelKey = 'default'): Promise<AnilibriaHistoryItem[]> {
+		if (this.hasLoadedHistory) {
+			this.hasReachedHistoryEnd = true;
+			return [];
+		}
+
+		const loggedIn = await this.checkLogin();
+		if (!loggedIn) {
+			this.hasReachedHistoryEnd = true;
```
MrMamen: Returning an empty list here makes a logged-out state indistinguishable from an empty history — the sync page will just show nothing. The other services throw (e.g. `'Invalid session'`) so the error surfaces to the user instead.

src/services/anilibria/AnilibriaApi.ts
```diff
@@ -0,0 +1,236 @@
...
+	name?: string | null;
+	name_english?: string | null;
+	ordinal?: number | null;
+	duration?: number | null;
+	updated_at?: string | null;
+	release_id: number;
+	release?: AnilibriaRelease | null;
+}
+
+interface AnilibriaRawTimecode extends Array<string | number | boolean | null> {
```
MrMamen: Minor: this would be simpler as a tuple type — `type AnilibriaRawTimecode = [string, number, boolean];` — instead of extending `Array` with indexed members.

MrMamen (changes requested):
Thanks for the contribution! The structure follows the service template well, the `isAbsolute` episode numbering matches the Crunchyroll precedent, the approximate watched-date limitation is honestly documented in both the service definition and the README, and lint/typecheck pass.

Both existing Copilot comments are correct and blocking, so please address them: the null-filtering in `convertHistoryItems()` breaks the positional 1:1 contract that `ServiceApi.loadHistory()` relies on (it indexes the converted array and calls `getDatabaseId()` on the result, so a filtered item shifts everything and crashes the sync), and the per-episode `Promise.all` fan-out is unbounded — a sequential loop like Netflix's `getHistoryMetadataFromSingleRequests()` would solve that. I've added a few additional comments inline.

dm3ch:
Addressed the review findings: AniLibria history conversion now preserves the shared positional contract and fetches episode metadata sequentially; session requests use the shared request layer with cancellation; the import timestamp cache is registered in typed storage and pruned; invalid sessions now surface as an error.

src/services/anilibria/AnilibriaApi.ts
```diff
@@ -0,0 +1,231 @@
...
+		return historyItem.releaseEpisodeId;
+	}
+
+	async convertHistoryItems(historyItems: AnilibriaHistoryItem[]): Promise<ScrobbleItem[]> {
+		const items: ScrobbleItem[] = [];
+		for (const historyItem of historyItems) {
+			const episode = await this.getEpisode(historyItem.releaseEpisodeId);
+			const item = this.convertEpisode(episode, historyItem.firstImportedAt, 100);
+			if (!item) {
+				throw new Error(`Could not convert AniLibria episode ${historyItem.releaseEpisodeId}`);
```
MrMamen, line 165: Non-blocking: throwing here means a single unconvertible episode (e.g. a release removed from AniLibria) aborts the whole history load. Preserving the contract this way is correct, but if you want to make it more resilient later, I did something similar in #493 for NRK by degrading to the data embedded in the history item instead of failing. Fine to leave as is for now.

MrMamen (approved):

PR #487, opened 2026-07-26
Support Netflix playback progress for history sync

smoothlystable:
I finally found a way to get proper progress (percentage watched) with Netflix, so not everything is assumed fully watched no matter how short you actually watched. Turns out the account metadata API (that we already use) holds that information, so the fix is quiet small. Seems like you can also get progress from their GraphQL API, but it didn't seem reliable enough to be worth it. 

Either way, I've tested it and it works great. I bet there are some issues that this should fix, but I'm too lazy right now to search for them.


MrMamen (approved):
Great! Thanks for your help with this!

PR #489, opened 2026-07-29
Add Apple TV web scrobbler

sitz:
## Summary

- add Apple TV as a scrobble-only service for `tv.apple.com`
- support on-demand movies and episodes launched from title pages, Continue Watching, featured content, homepage shelves, and autoplay transitions
- exclude live sports, trailers, extras, previews, and promotional/background videos
- list Apple TV in the existing README service table

## Why

Apple TV often mounts the real player as an overlay without changing the homepage URL. Route-only detection therefore misses playback launched from several homepage surfaces and can retain stale metadata during SPA or autoplay transitions.

The parser combines exact content-route and Apple metadata validation with scoped player hooks. When the URL remains on the homepage, it resolves matching Apple content IDs from the page's content links and fingerprints player metadata so item changes invalidate the cache.

## Implementation

- register the `apple-tv` service for `*://*.tv.apple.com/*`, with scrobbling enabled and sync disabled
- track only `[data-testid="video-container"] video`
- parse locale-independent movie, episode, and show routes using exact `umc.cmc.*` IDs
- parse validated Movie and Episode JSON-LD where available
- use player title/subtitle metadata for homepage overlays, show-page playback, SPA navigation, and autoplay
- cross-check movie player titles against page metadata so trailers and extras cannot inherit a parent movie ID
- add Apple TV to the README service table without reformatting it

## Validation

- `pnpm tsc`
- `pnpm biome check src/services/apple-tv README.md`
- `pnpm run build-dev`
- verified `apple-tv.js` in Chrome and Firefox builds
- verified Apple TV optional host permission in both generated manifests
- inspected authenticated live movie and episode playback from featured content, Continue Watching, standard shelves, and button-only homepage cards

### DIFF
--- a/src/services/apple-tv/AppleTvParser.ts
+++ b/src/services/apple-tv/AppleTvParser.ts
@@ -42,10 +42,11 @@ interface AppleTvEpisodeDetails {
 /**
  * Apple TV renders promotional videos throughout the site, while the actual long-form
  * player is mounted in a dedicated `video-container`. Item metadata is server-rendered
- * as semantic meta tags and JSON-LD, while series-page playback exposes the current
- * episode through player metadata and the episode shelf. Page metadata is checked
- * against the current `umc.cmc.*` URL ID, and player metadata is fingerprinted so an
- * autoplay transition cannot retain the previous episode.
+ * as semantic meta tags and JSON-LD, while active player overlays expose the current
+ * item through player metadata. Homepage playback can leave the URL at its locale root,
+ * so matching content links provide exact IDs when available. Page metadata is checked
+ * against the current `umc.cmc.*` URL ID, and player metadata is fingerprinted so SPA
+ * and autoplay transitions cannot retain the previous item.
  */
 class _AppleTvParser extends ScrobbleParser {
 	private itemIdentity: string | null = null;
--- a/src/services/apple-tv/AppleTvParser.ts
+++ b/src/services/apple-tv/AppleTvParser.ts
@@ -39,6 +39,16 @@ interface AppleTvEpisodeDetails {
 	title: string;
 }
 
+interface AppleTvPlayerMetadata {
+	subtitle: string;
+	title: string;
+}
+
+interface AppleTvPlaybackIdentityCache {
+	identity: string;
+	key: string;
+}
+
 /**
  * Apple TV renders promotional videos throughout the site, while the actual long-form
  * player is mounted in a dedicated `video-container`. Item metadata is server-rendered
@@ -51,6 +61,7 @@ interface AppleTvEpisodeDetails {
 class _AppleTvParser extends ScrobbleParser {
 	private completionPending = false;
 	private itemIdentity: string | null = null;
+	private playbackIdentityCache: AppleTvPlaybackIdentityCache | null = null;
 
 	constructor() {
 		super(AppleTvApi, {
@@ -87,6 +98,7 @@ class _AppleTvParser extends ScrobbleParser {
 		super.clearItem();
 		this.completionPending = false;
 		this.itemIdentity = null;
+		this.playbackIdentityCache = null;
 	}
 
 	protected parseItemFromApi(): Promise<ScrobbleItem | null> {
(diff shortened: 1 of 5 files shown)


src/services/apple-tv/AppleTvParser.ts
```diff
@@ -0,0 +1,491 @@
...
+			anchor.textContent?.trim() ?? '',
+		].filter(Boolean);
+	}
+
+	private normalizeTitle(value: string): string {
+		return value
+			.toLocaleLowerCase()
+			.replace(/[^\p{L}\p{N}]+/gu, ' ')
+			.trim();
+	}
```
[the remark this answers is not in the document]
sitz: Addressed in f3a6dd8: title normalization now uses locale-independent `toLowerCase()` for deterministic matching.

src/services/apple-tv/AppleTvParser.ts
```diff
@@ -0,0 +1,491 @@
...
+	private getMetaContent(name: string, attribute = 'name'): string {
+		return (
+			document.querySelector<HTMLMetaElement>(`meta[${attribute}="${name}"]`)?.content?.trim() ?? ''
+		);
+	}
+
+	private parseNonNegativeInteger(value: number | string | undefined): number | null {
+		const parsed = typeof value === 'number' ? value : parseInt(value ?? '');
+		return Number.isInteger(parsed) && parsed >= 0 ? parsed : null;
+	}
```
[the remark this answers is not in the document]
sitz: Addressed in f3a6dd8: added radix 10 to JSON-LD number parsing, and to movie year parsing for consistency.

src/services/apple-tv/AppleTvParser.ts
```diff
@@ -0,0 +1,491 @@
...
+		const pageTitle = this.getMetaContent('apple:title');
+		const playerTitle = document
+			.querySelector('[data-testid="video-player"] [data-testid="player-metadata-title"]')
+			?.textContent?.trim();
+		return (
+			!!pageTitle &&
+			!!playerTitle &&
+			this.normalizeTitle(pageTitle) === this.normalizeTitle(playerTitle)
+		);
+	}
```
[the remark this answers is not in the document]
sitz: Addressed in f3a6dd8: movie parsing and the trailer/extras guard now share the same title fallback chain: `apple:title`, exact-content-ID-validated Movie JSON-LD, then `og:title`.

MrMamen (commented):
This is a well-built parser — I verified the tricky base-class interactions against `ScrobbleEvents`/`ScrobbleController` (the empty-playback tick → `stopScrobble()` → `clearItem()` flow that resets `completionPending`/`itemIdentity` works exactly as the code comments describe), the `parseItemFromApi()` override correctly avoids the base class's unimplemented `getItem()` path, and the scrobble-only empty API class is fine since scrobbling doesn't require a service login. The earlier Copilot comments are all addressed in f3a6dd8. Nothing blocking from my side — a couple of non-blocking suggestions inline.

src/services/apple-tv/AppleTvParser.ts
```diff
@@ -0,0 +1,500 @@
...
+		const parsedShowTitle = subtitleMatch ? showTitle : subtitle;
+
+		if (!episodeTitle || !parsedShowTitle) {
+			return null;
+		}
+
+		return { season, number, title: episodeTitle, showTitle: parsedShowTitle };
+	}
+
+	private getPlaybackIdentity(route: AppleTvRoute | null): string | null {
```
MrMamen, line 274: Non-blocking: `parsePlayback()` runs every 0.5s (`checkFrequency`), and this recomputes the identity each tick — including `findContentId()`, which scans and normalizes every matching anchor in the document. On a homepage with hundreds of shelf cards that's a fair amount of repeated work. Caching the computed identity keyed on the raw title/subtitle text (and recomputing only when they change) would make the steady-state tick nearly free.

src/services/apple-tv/AppleTvParser.ts
```diff
@@ -0,0 +1,500 @@
...
+			this.getShowTitleFromLdName(ld.name, title);
+
+		if (season === null || number === null || !title || !showTitle) {
+			return null;
+		}
+
+		return { season, number, title, showTitle };
+	}
+
+	private parseItemFromPlayer(route: AppleTvRoute | null): ScrobbleItem | null {
```
MrMamen, line 207: Nit: the player title/subtitle extraction here is duplicated in `getPlaybackIdentity()` — could be a small shared helper returning `{ title, subtitle }`.

MrMamen (approved):

src/services/apple-tv/AppleTvParser.ts
sitz: Addressed in 29438ff. Playback identity is now cached by route ID/type plus raw player title/subtitle, so steady-state 500 ms ticks avoid shelf scans. The cache is cleared with the item, and misses remain uncached so late-loading homepage shelves can still resolve.

src/services/apple-tv/AppleTvParser.ts
sitz: Addressed in 29438ff. Added a shared `getPlayerMetadata()` helper and reused it for item parsing, identity calculation, and movie title validation.

PR #494, opened 2026-08-13
Verify Trakt search results and add slug/TMDB ID fallbacks for shows

MrMamen:
## Problem

Trakt's text search can return confidently wrong results, which then get cached and used for every scrobble/sync:

- Searching for the Norwegian show **"Åsted Norge"** returns exactly one result: *Crime Scene Berlin: Nightlife Killer* (matched through its Norwegian translated title) — while the actual show, which exists on Trakt with an exactly matching title (`asted-norge`, trakt 120617), isn't returned by the text search at all. This appears to be a search index gap on Trakt's side and affects many shows (verified: Klassen, VGS, Dagsrevyen, Dyrepasserne, Familien Lykke, Superkrim). No search parameter helps — `fields=title` still leaks translation matches and still doesn't return the actual show.
- `findItem()` used a single search result blindly, skipping both the exact-title verification and the TMDB cross-reference, so the wrong show was cached and every episode lookup 404'd.

## Fix

The matching chain for shows is now: **exact title/year match among the search results → slug lookup → TMDB cross-reference → first search result** (unchanged last resort).

- Search results are always verified against title/year, even when there is only one.
- New slug lookup: Trakt slugs are transliterated titles (`Åsted Norge` → `asted-norge`), so a direct `GET /shows/{slug}` finds shows the text search index misses, at the cost of a single request. The result is only accepted when its title transliterates back to the same slug and the year matches (when known), so colliding slugs for different shows are rejected.
- The TMDB cross-reference no longer requires the show to be among the text search results: when it isn't, the show is looked up on Trakt directly by its TMDB ID (`GET /search/tmdb/{id}?type=show`).

## Validation

Tested with cleared caches against real history syncs from five services (NRK, TV 2 Play, Viaplay, Netflix + scrobbling):

- **Slug lookup: 7 correct rescues, 0 false accepts** (Åsted Norge, Klassen, VGS, Dagsrevyen, Dyrepasserne, Familien Lykke, Superkrim — all missing from their own title searches).
- **TMDB step: 4 correct rescues, 0 false accepts** (Tuva og Ronnys Bed & Breakfast → *Tuva and Ronny's Bed & Breakfast*, Farfar → *Granddad*, Harry Hole → *Detective Hole*, Norske Krimsaker → *Norwegian Crime Stories* — all listed under English titles on Trakt).
- The safety guard proved itself on *The Blacklist* via Viaplay: the service supplies per-season production years (2019/2020), so the slug result (2013) was rejected by the year check — and the TMDB step still resolved the correct show downstream.
- Exact matches cut through even 37–75 search results (Taken, Rush), so the chain adds no cost in the normal case.
- Content that genuinely isn't on Trakt (sports broadcasts, small web shows) still falls through to the first result or a clean 404, as before.

Developed in collaboration with Claude AI.

### DIFF
--- a/src/apis/TraktSearch.ts
+++ b/src/apis/TraktSearch.ts
@@ -264,8 +264,12 @@ class _TraktSearch extends TraktApi {
 			if (matchesTitle && matchesYear) {
 				return { show };
 			}
-		} catch (_err) {
-			// No show with this slug
+		} catch (err) {
+			if (!(err instanceof RequestError) || err.status !== 404) {
+				// Only a 404 means there is no show with this slug - propagate cancellations
+				// and other failures instead of falling through to a first-result guess
+				throw err;
+			}
 		}
 		return undefined;
 	}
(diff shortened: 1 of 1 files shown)


MrMamen:
@sitz @smoothlystable @Superritchman Given your recent contributions I invite you to test and/or review these changes as they affect all streaming services.

Superritchman:
I tested Crunchyroll and Amazon against my own history.

**Crunchyroll**
Couldn't find any calls against TMDB (aside from image requests). Looks like the Trakt slug matching has a great hit rate for my test items (~30 shows).
Every show was matched correctly. Only single episodes were unmatched. (Hence Crunchyroll's "unique" season grouping.)

**Amazon**
Could only test 4 shows, but one triggered the fallback and it worked as intended (the-legend-of-vox-machina).
Every show was matched correctly.

Minor note: I had to switch to `Bearer Token` authentication for TMDB. Somehow, via the `api_key` param, I got an invalid API key message (401).

MrMamen:
> Minor note: I had to switch to `Bearer Token` authentication for TMDB. Somehow, via the `api_key` param, I got an invalid API key message (401).

Did you remember to get API-key from here: https://www.themoviedb.org/settings/api and but it into your env-files?

Anyways remember to delete cache from the settings in the extension if you want to test, as results are cached. Thanks anyway

smoothlystable:
It seems to work great in general, but I think I found a related bug for Viaplay.  
Viaplay gets year for episodes from episode release year and not show release year. So a show that was released 2020 with an episode that was released 2022, it uses 2022 and with this new stricter search it tries to find a match for show title + 2022 which doesn't exist.

I can help you implement a fix if you want. Not sure if you can give me push access to only this PR branch, otherwise I can open another PR if you prefer?

smoothlystable:
I created a new PR #495 as I found another bug as well and the fixes can be applied cleanly to the codebase without this PR anyway.

MrMamen:
> Not sure if you can give me push access to only this PR branch, otherwise I can open another PR if you prefer?
Not sure if it's possible. But if you find anything else related to this branch you can fork it and create a PR set to merge into this branch instead of master. I've rebased your fix into this now.

Superritchman:
> > Minor note: I had to switch to `Bearer Token` authentication for TMDB. Somehow, via the `api_key` param, I got an invalid API key message (401).
> 
> Did you remember to get API-key from here: https://www.themoviedb.org/settings/api and but it into your env-files?
> 
> Anyways remember to delete cache from the settings in the extension if you want to test, as results are cached. Thanks anyway

Found my mistake: I was using the newish read only API token. These keys only work with auth via bearer header. 

~I checked Crunchyroll and Amazon again and could verify the same behaviour as @smoothlystable.~

After some clarifications within the proposed Crunchyroll fix, I rechecked all services again. Every service just sets `EpisodeItem.year`, but not the nested `EpisodeItem.show.year`.

PR #495, opened 2026-08-15
Fix Viaplay history sync

smoothlystable:
After clearing cache while testing PR #494 I found two bugs in the Viaplay adapter;
- For TV show episodes it sets show year to the year the episode was released (which not always matches show release year).
- It couldn't handle the end of history when it's multiple pages.


MrMamen (approved):
Great fixes. Thanks for testing #494 so thoroughly!

PR #496, opened 2026-08-17
Crunchyroll season matching

Superritchman:
Omit the episode release date to support stricter slug/tmdb matching (see #494).

Also improved season matching by comparing against the slug. Crunchyroll often adds OVAs and related movies as separate seasons, so the real season numbering is often wrong. Season slug (if present) usually holds the real season name, which is most often "season-<number>".

Let me know if I should split the commits into two PRs.

### DIFF
--- a/src/services/crunchyroll/CrunchyrollApi.ts
+++ b/src/services/crunchyroll/CrunchyrollApi.ts
@@ -33,19 +33,22 @@ export interface CrunchyrollHistoryItem {
 	parent_type?: string;
 	panel: {
 		title: string;
-		episode_metadata: {
-			series_id: string;
-			season_number: number;
-			season_id: string;
-			episode: string;
-			episode_number?: number;
-			sequence_number?: number;
-			episode_air_date: Date;
-			season_title: string;
-			series_title: string;
-			duration_ms: number;
-		};
+		episode_metadata: CrunchyrollEpisodeMetadata;
 	};
+}
+
+export interface CrunchyrollEpisodeMetadata {
+	series_id: string;
+	season_number: number;
+	season_id: string;
+	episode: string;
+	episode_number?: number;
+	sequence_number?: number;
+	episode_air_date: Date;
+	season_title: string;
+	season_slug_title?: string;
+	series_title: string;
+	duration_ms: number;
 }
 
 class _CrunchyrollApi extends ServiceApi {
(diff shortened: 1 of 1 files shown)


src/services/crunchyroll/CrunchyrollApi.ts
```diff
@@ -198,11 +201,15 @@ class _CrunchyrollApi extends ServiceApi {
...
-					season: metadata.season_number || 0,
+					// season numbering is often not aligned with official seasons
+					// check against slug to receive more aligned season number
+					season: this.getBestSeasonNumber(metadata),
 					// Crunchyroll often numbers anime episodes sequentially across all seasons,
 					// so allow episode matching to resolve the number as absolute.
 					isAbsolute: true,
-					year: new Date(metadata.episode_air_date).getUTCFullYear(),
+					// `metadata.episode_air_date` is the episode's release year, not the
+					// series premiere year. Leave the show year unknown so it cannot reject
```
MrMamen: This rationale applies to #495 (Viaplay), where the year sat on the **show** object — but here the removed year is on the **episode** item, which isn't consulted during show matching at all: `findShow()` passes `item.show` to `findItem()`, and `tryTmdbFallback()` uses `item.show.year` — and Crunchyroll's show object (below) has never had a year. So this commit only removes the episode's year metadata without any matching benefit, and the comment would mislead future readers. I'd suggest dropping this commit / restoring the year — unless you've observed an actual mis-match caused by it, in which case I'd love to see the case.

src/services/crunchyroll/CrunchyrollApi.ts
```diff
@@ -220,6 +227,22 @@ class _CrunchyrollApi extends ServiceApi {
 		return Promise.resolve(items);
 	}
 
+	getBestSeasonNumber(metadata: CrunchyrollEpisodeMetadata): number {
```
MrMamen, line 228: Minor: Crunchyroll also lists movies/specials as separate "seasons" — if those have recognizable slugs (e.g. `-movie`), the same treatment as OVAs might apply. Fine as a follow-up.

MrMamen (changes requested):
The season-from-slug commit is a solid improvement — the OVA → season 0 (Trakt specials) mapping and the conservative `season-N` extraction with fallback both make sense, and the metadata type extraction is clean. The year commit, however, doesn't do what its comment says for this service — see inline. To answer your question: no need to split, one PR is fine (especially since the year commit likely just gets dropped).

src/services/crunchyroll/CrunchyrollApi.ts
Superritchman: I see, I will drop the commit. Didn't saw that the viaplay "year" was part of the nested `show` property.

MrMamen (approved):

PR #498, opened 2026-08-19
Crunchyroll movie matching

Superritchman:
Check for `movie(s)` in series/season slug for increased movie matching. Movie titles are sometimes only present as season or show title.
If this matching still misses some movies, it could be matched against localized versions of `movie` in some future versions.

Also add thumbnails for the matching preview.

### DIFF
--- a/src/services/crunchyroll/CrunchyrollApi.ts
+++ b/src/services/crunchyroll/CrunchyrollApi.ts
@@ -34,6 +34,9 @@ export interface CrunchyrollHistoryItem {
 	panel: {
 		title: string;
 		episode_metadata: CrunchyrollEpisodeMetadata;
+		images?: {
+			thumbnail?: CrunchyrollThumbnail[][];
+		};
 	};
 }
 
@@ -50,6 +53,13 @@ export interface CrunchyrollEpisodeMetadata {
 	series_title: string;
 	series_slug_title?: string;
 	duration_ms: number;
+}
+
+export interface CrunchyrollThumbnail {
+	height: number;
+	source: string;
+	type: string;
+	width: number;
 }
 
 class _CrunchyrollApi extends ServiceApi {
@@ -177,12 +187,16 @@ class _CrunchyrollApi extends ServiceApi {
 		for (const historyItem of historyItems) {
 			const metadata = historyItem.panel.episode_metadata;
 			const title = this.getNormalizedTitle(historyItem);
+			const thumbnail = historyItem.panel.images?.thumbnail
+				?.find((o) => o.length > 0)
+				?.find((t) => !!t.source);
 
 			if (this.isMovie(historyItem)) {
 				const item = new MovieItem({
 					id: historyItem.id,
 					serviceId: this.id,
 					title: title,
+					imageUrl: thumbnail?.source,
 					year: new Date(metadata.episode_air_date).getUTCFullYear(),
 					watchedAt: Utils.unix(historyItem.date_played),
 					progress: historyItem.fully_watched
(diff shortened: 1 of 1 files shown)


src/services/crunchyroll/CrunchyrollApi.ts
```diff
@@ -61,6 +72,9 @@ class _CrunchyrollApi extends ServiceApi {
 	isActivated: boolean;
 	session: CrunchyrollSession | null = null;
 
+	movieRegex: RegExp = /(?:^|-)movies?|mov(?:-|$)/i;
```
MrMamen: The alternation splits as `(?:^|-)movies?` OR `mov(?:-|$)` — the first alternative has no right boundary (a slug like `die-moviestars` classifies as a movie) and the second has no left boundary (`kamov-47` too). Assuming the intent is to match a whole slug segment: `/(?:^|-)(?:movies?|mov)(?:-|$)/i`.

src/services/crunchyroll/CrunchyrollApi.ts
```diff
@@ -241,6 +251,32 @@ class _CrunchyrollApi extends ServiceApi {
...
+
+	getNormalizedTitle(historyItem: CrunchyrollHistoryItem) {
+		const title = historyItem.panel.title.replace(this.dubSubSuffix, '');
+		const metadata = historyItem.panel.episode_metadata;
+		if (title.toLowerCase() === 'episode') {
+			return `${title} ${metadata.episode || metadata.sequence_number}`;
+		}
+		if (title.toLowerCase() === 'movie' || title.trim().length === 0) {
+			return (
+				metadata.season_title.replace(this.dubSubSuffix, '') ||
```
MrMamen, line 273: `isMovie()` guards `season_title` with `!!`, but this line calls `.replace()` on it unguarded — so if the field can actually be missing, the title normalization throws for exactly the movie items it's meant to rescue. Suggest `metadata.season_title?.replace(...) || metadata.series_title?.replace(...) || title`.

src/services/crunchyroll/CrunchyrollApi.ts
```diff
@@ -241,6 +251,32 @@ class _CrunchyrollApi extends ServiceApi {
 		return Number.parseInt(season, 10);
 	}
 
+	isMovie(historyItem: CrunchyrollHistoryItem): boolean {
+		const metadata = historyItem.panel.episode_metadata;
+		return (
+			(!!metadata.season_title && metadata.season_title.toLowerCase().includes('movie')) ||
+			historyItem.panel.title.toLowerCase() === 'movie' ||
+			metadata.episode_number === null ||
```
MrMamen, line 259: Following up on the Copilot comment, this is worth answering with data: do Crunchyroll movie entries always carry an explicit `"episode_number": null`, or can the field be absent? If it can be absent, `== null` (loose) would cover both — but note the flip side: any episodes/OVAs *without* numbers would then classify as movies, undoing the season 0 mapping from #496. If your data shows movies always have an explicit null, the strict check is actually the safer choice and just deserves a short comment saying so.

MrMamen (changes requested):
Nice follow-up on the movies-as-seasons thread from #496 — the title fallback to season/series title is exactly what those items need. A few things inline before merge: the movie regex has asymmetric anchoring with real false-positive potential, `getNormalizedTitle()` can throw for the exact items it's meant to handle, and the Copilot comment about `episode_number` deserves an answer from your data (both the strict and the loose null check have failure modes here).

src/services/crunchyroll/CrunchyrollApi.ts
Superritchman: Valid concern, fixed it with your proposed regex.

Superritchman:
I checked through my history (~4,300 entries). Although the data itself was inconsistent, the data structure was very stable. No optional parameters other than the whole panel object being missing when the show was deleted from CR.

Therefore, I removed the optional flag from most of the JSON properties (except thumbnail), and removed some null/undefined checks for the same reason. Properties are always present, and all title/slug fields are served with text or an empty string.

So `"episode_number"` will always be present, either filled with a `number` or `null`.

MrMamen (approved):
seems like the type on episode_number could be numer|null instead?

